PR #38, opened 2020-07-23
New tests and scripts

dr-rodriguez:
Adding a test to verify the integrity of the Names table (checks that each Source has an entry there and that there is at least one valid entry). 
Also adds/updates some of the example scripts I've been using for loading BDNYC data.
Finally, switched back to using the github install of astrodbkit2 so we can use new features prior to official releases.


.github/workflows/python-app.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install flake8 pytest
         # install dependencies. Eventually, this should all be in in requirements.txt 
-        pip install astrodbkit2
+        pip install git+https://github.com/dr-rodriguez/AstrodbKit2
```
kelle, line 28: I dont understand why we arent using the pip version?

dr-rodriguez: By changing to the github version we can capture the latest work, as opposed to released versions

PR #40, opened 2020-08-06
Cleaning up duplicate test

dr-rodriguez:
Cleaning up a duplicated test in test_integrity.py

### DIFF
--- a/tests/test_integrity.py
+++ b/tests/test_integrity.py
@@ -99,7 +99,7 @@ def test_source_names(db):
     assert len(missing_names) == 0
 
 
-def test_source_uniqueness1(db):
+def test_source_uniqueness(db):
     # Verify that all Sources.source values are unique
     source_names = db.query(db.Sources.c.source).astropy()
     unique_source_names = unique(source_names)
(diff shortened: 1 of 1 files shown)


kelle:
Even though these two tests kind of do the same thing, they do it differently. (When I wrote them, I remember thinking that they are technically different.)  I don't mind having extra tests since the different approaches could theoretically find weird things. Is there any good reason to delete it? It's not very resource intensive.

dr-rodriguez:
My main reason was to clean up duplicated code. At some point we may have tons of tests and having two (or more) that do the same thing is probably not idea. At the moment, you're right- it's not particularly resource intensive. 
An alternative could be to have the two independent checks being handled by the same test_source_uniqueness function (ie, two assert statements in that single call).

kelle:
I like the idea of two asserts in one test! I didn't know you could do that. I guess both asserts need to pass in order for the test to pass?

dr-rodriguez:
Correct. Though if the first assert fails, any subsequent ones are not checked which is one drawback of that setup.

kelle:
I like that compromise. Can you revise the PR?

Issue #20, opened 2020-07-10
Provide sample duplication script


dr-rodriguez:
As sources are ingested, there is a possibility that we end up with duplicated sources. We should write a sample script that examines the database and identifies potential duplicates. This can also fold into our integrity tests.

kelle:
Is this addressed by the tests in #36 ?

dr-rodriguez:
We need a more complex test. 
What I'd like to do is go through every source, gather alternate designations for each, and then check against other_names in the Names table to verify that indeed we don't list the same source twice under different designations.

I have an idea (https://github.com/dr-rodriguez/SIMPLE/commit/e12a099e4aae41c3a293dcd4550b84e9fd2b4508), though a brief check showed that it times out against Simbad (too many queries).

kelle:
I implemented this in the sort_sources function in útils.

PR #41, opened 2020-08-07
Adding test with Simbad check

dr-rodriguez:
This adds a test to address issue #20. It will query Simbad for all sources by their Source.source name and then loop over the results to verify that they are not present in the database for more than one source. It handles the Simbad query properly to avoid hundreds of individual calls to their service.

Matching is done by exact name, so binaries will generally not match against each other (ie, TWA 27 does not match TWA 27B).


tests/test_integrity.py
```diff
@@ -127,6 +129,39 @@ def test_source_uniqueness2(db):
...
+    # Query Simbad and confirm that there are no duplicates with different names
+
+    # Get list of all source names
+    results = db.query(db.Sources.c.source).all()
+    name_list = [s[0] for s in results]
+
+    # Add all IDS to the Simbad output as well as the user-provided id
+    Simbad.add_votable_fields('ids')
+    Simbad.add_votable_fields('typed_id')
+    res = Simbad.query_objects(name_list)
```
kelle, line 142: consider renaming `res` to `simbad_results`

dr-rodriguez:
Correct. 
The example I was hoping to catch was TWA 27, which was listed as two independent sources: 
TWA 27 and 2MASS J12073346-3932539
Instead, they should have been two names for the same single source.

kelle:
Great! Happy to merge once conflict is resolved.

PR #44, opened 2020-08-07
add project overview to the README

kelle:


README.md
```diff
@@ -1,5 +1,9 @@
 # SIMPLE
 
+The vision of the SIMPLE project is to create a *collaborative* database of low-mass stars, brown dwarfs, and directly imaged exoplanets: a simple archive of complex objects. We plan to include coordinates, photometry, kinematics, spectra, images, as well as modeled and retrieved parameters. We are developing methods to interact with the database via python, a website, and database browsers.
```
dr-rodriguez, line 3: Looks good, but I notice you mention modeled and retrieved parameters which is still an open question. Do you think this commits us to supporting this?

Also, I would say: 
We are developing tools to interact with the database via several methods, including python, a website and API, and database browsers.

PR #54, opened 2020-08-18
Some minor debug messages when creating database

dr-rodriguez:
Some minor debug messages when creating database


dr-rodriguez:
Updating to alter the integrity test Action to not fire on `push` but have `workflow_dispatch` This will allow us to run it manually and will not automatically run when there are pushes to the main branch. This wont happen anyways as this is a protected branch, but may be blocking other Actions from running.

PR #57, opened 2020-08-21
Adding Parallaxes table

dr-rodriguez:
Now that a sample of sources has been ingested and the API is somewhat set, I think it's time to finalize the data model.
Over the next few pull requests I hope to add tables addressing the rest of the schema.

In this one, I add a Parallaxes table along with an appropriate test for it.


tests/test_integrity.py
```diff
@@ -216,6 +216,22 @@ def test_photometry(db):
     assert len(t) == 0
 
 
+def test_parallaxes(db):
```
Will-Cooper, line 219: will be overwritten by #59

PR #58, opened 2020-08-21
Adding ProperMotions table

dr-rodriguez:
Adding a ProperMotions table to store these values in the database. 
Should be merged after the Parallaxes table PR is merged (#57).


tests/test_integrity.py
```diff
@@ -216,6 +216,36 @@ def test_photometry(db):
     assert len(t) == 0
 
 
+def test_parallaxes(db):
```
Will-Cooper, line 219: will be overwritten by #59

Issue #61, opened 2020-08-25
Storyboard/ mockup of front end


Will-Cooper:
Make a mockup of how we want the front end web page to look, starting from exoMAST and what they have. Consider what we need/ would like.

PR #62, opened 2020-08-26
Add search functionality to the same page

Will-Cooper:
Stopped bokeh warning about column data source columns.
Created callback enabled front facing div to hide/ unhide with the search functionality on
Searches using db.search_object so partial strings
That will update the main table on key press; on non-focus to search box (enter or click away) the search div will vanish, it'll come back if you click the button on the top.

Still ugly css as haven't had the time to play with that too much. Should be able to style bokeh elements and surroundings. Jinja2 still messing up the css.

Still call as ```bokeh serve --show website/``` from top directory.

### DIFF
--- a/website/main.py
+++ b/website/main.py
@@ -12,6 +12,8 @@ spectra_fig
 main
     Generating initial elements on current document and assigns callback to fill them on selection
 """
+# inbuilt libs
+from string import Template
 # 3rd party imports
 from bokeh.models import ColumnDataSource, PreText, DataTable, TableColumn, TextInput, CustomJS, Button
 from bokeh.plotting import curdoc, figure
@@ -19,8 +21,10 @@ import numpy as np
 # local imports
 import splash_table as st
 import gen_info as geninfo
-import gen_plot as genspec
+#  import gen_plot as genspec
 import gen_photo as genphot
+
+input_string = Template('$typed')
 
 
 def info_box():
@@ -150,7 +154,12 @@ def main():
         new
             The new object (required under hood)
         """
-        selected_index = dtcds.selected.indices[0]  # the index selected from datatable on page
+        for _ in (attrname, old, new):
+            pass  # make pycharm shut up about unused parameters
+        try:
+            selected_index = dtcds.selected.indices[0]  # the index selected from datatable on page
+        except IndexError:
+            return
         data = dtcds.data  # the data in table
         target = data['sname'][selected_index]  # pulling out object name
 
@@ -162,6 +171,8 @@ def main():
 
         specfig.title.text = target  # change title on spectra to object name
         # TODO: add spectra in here
+        sbar.value = 'bye'
+        sbar.value = ''
         return
 
     def sbar_update(attrname, old, new):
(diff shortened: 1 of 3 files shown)


website/main.py
```diff
@@ -129,16 +155,39 @@ def update(attrname, old, new):
         target = data['sname'][selected_index]  # pulling out object name
 
         band, mags, magserr = genphot.main(target)  # grab photometric data fro given object
-        photcds.data['bands'] = band  # update photometric table with known bands
-        photcds.data['vals'] = mags  # update photometric table with corresponding magnitudes
-        photcds.data['valserr'] = magserr  # update photometric table with corresponding magnitude errors
+        photcds.data = dict(bands=band, vals=mags, valserr=magserr)
```
Will-Cooper, line 158: couldn't remember which sources actually have photometry to check this is fine but it should be (and also stop a warning message on terminal); use identical method later which works

website/splash_table.py
```diff
@@ -53,6 +63,29 @@ def make_table(results: list):
...
+    -------
+    The list of tuples of object info
+    """
+    qry = Qry(qrystr)
+    db = load_db()  # load the database
+    results = db.search_object(qrystr)
+    if qry.default:
+        return results
+    else:
+        results = [obj[:3] for obj in results]
```
Will-Cooper, line 85: cuts to name, ra, dec

website/templates/index.html
```diff
@@ -34,14 +34,6 @@
     <!-- Font Awesome v.5.0.12 -->
     <link rel="stylesheet" href="/static/css/all.css" crossorigin="anonymous">
 
-    <!-- jquery datatable -->
```
Will-Cooper, line 37: datatable wasn't working and shouldn't actually be necessary to have multiple pages if it's now fed by a search

dr-rodriguez:
Quickly tested and I'm seeing duplicated photometry values. Try TWA 26 and you'll see two entries for each WISE band, for example.

website/splash_table.py
```diff
@@ -53,6 +63,29 @@ def make_table(results: list):
     return cds, dt
 
 
+def querying(qrystr: str = 'notanobject'):
```
dr-rodriguez, line 66: I'd recommend using some more descriptive function and variable names. You have querying, qry, Qry, qrystr all here. A longer variable name would go a long way to make the code more readable.

website/splash_table.py
dr-rodriguez: Remember that you can request search_object to return data in a different format, in case astropy Tables or pandas Dataframes would be easier for you to work with.

Also, this comment 'cuts to name, ra, dec' should be explicitly in the code

website/main.py
```diff
@@ -129,16 +155,39 @@ def update(attrname, old, new):
...
+            The new object (required under hood)
+        """
+        qry = sbar.value_input
+        results = np.array(st.querying(qry))
+        snames, ras, decs = [], [], []
+        if results.size:  # not empty
+            snames = results[:, 0]
+            ras = results[:, 1]
+            decs = results[:, 2]
+        dtcds.data = dict(sname=snames, ra=ras, dec=decs)
```
dr-rodriguez, line 188: There may be ways to simplify this re-formatting of the data, depending on which format you request the results as.

dr-rodriguez:
The initial search panel window is huge in my screen, so I have to scroll a bit to get to the results table. Also, the results aren't cleared when I get back to that page which means I can click them without anything happening. I have to make a completely different search to get responsive results.

Overall, this looks good, but we've lost the ability to browse through the full table of sources.
I would suggest that more design work be done first (as part of #61) before continuing to iterate on what you have, with the exception of fixing bugs and such.

Will-Cooper:
> The initial search panel window is huge in my screen, so I have to scroll a bit to get to the results table. 

That's odd, sounds like I need to do more testing on the front end in different browsers/ resolutions (CSS is a nightmare sometimes...).

> Also, the results aren't cleared when I get back to that page which means I can click them without anything happening. I have to make a completely different search to get responsive results.

I can add that fairly simply
> Overall, this looks good, but we've lost the ability to browse through the full table of sources.

An empty search _should_ provide the full table but I'll ensure it does

website/main.py
Will-Cooper: It's fine, bar the issue with duplicates in #64

website/main.py
```diff
@@ -12,16 +12,20 @@
...
 from bokeh.plotting import curdoc, figure
 import numpy as np
 # local imports
 import splash_table as st
 import gen_info as geninfo
-import gen_plot as genspec
+#  import gen_plot as genspec
 import gen_photo as genphot
 
+input_string = Template('$typed')
```
Will-Cooper, line 27: close attack vector (technically possible without templated strings)

website/templates/index.html
```diff
@@ -117,15 +117,24 @@ <h1 style="color: floralwhite; position: relative; top: 45%; left: 35% ; width:
...
-            {%- endblock -%}
+        <div id="searchWrapper" style="position: absolute; top: 110px; width: 100%; text-align: center;
+         height: 740px; background-color: darkslateblue; overflow-y: hidden; z-index: 3; display: block">
+            <div id="searchBar" style="display: inline-block; width: 40%; top: 45%; font-size: large; position: relative">
+                {%- block searchbar -%}
+                {{ embed(roots.searchbar) }}
+                {%- endblock -%}
+            </div>
         </div>
+        <script>
```
Will-Cooper, line 129: every 2 seconds makes the page focus on the search bar if its on screen

website/splash_table.py
```diff
@@ -63,27 +72,22 @@ def make_table(results: list):
...
     """
-    qry = Qry(qrystr)
     db = load_db()  # load the database
-    results = db.search_object(qrystr)
-    if qry.default:
-        return results
-    else:
-        results = [obj[:3] for obj in results]
-        return results
+    results = db.search_object(query_string, format='pandas')
```
Will-Cooper, line 89: using pandas dataframe instead of list for enhanced readability and general betterness

dr-rodriguez (approved):

Issue #22, opened 2020-07-10
Add Gravities table


dr-rodriguez:
We should add a table to store gravity information. Here is the original prototype for reference:
```
class Gravities(Base):
    __tablename__ = 'Gravities'
    source = Column(String(100), ForeignKey('Sources.source'), nullable=False, primary_key=True)
    gravity = Column(Enum(Gravity))  # restricts to enumerated values
    regime = Column(Enum(Regime), primary_key=True)  # restricts to a few values: Optical, Infrared
    comments = Column(String(1000))
    reference = Column(String(30), ForeignKey('Publications.name'), primary_key=True)
```

Issue #21, opened 2020-07-10
Add SpectralTypes table


dr-rodriguez:
We should have a table for storing spectral types. Here was the prototype, which may need some modifications:
```
class SpectralTypes(Base):
    __tablename__ = 'SpectralTypes'
    source = Column(String(100), ForeignKey('Sources.source'), nullable=False, primary_key=True)
    spectral_type = Column(Float)
    spectral_type_error = Column(Float)
    regime = Column(Enum(Regime), primary_key=True)  # restricts to a few values: Optical, Infrared
    best = Column(Boolean)  # flag for indicating if this is the best measurement or not
    comments = Column(String(1000))
    reference = Column(String(30), ForeignKey('Publications.name'), primary_key=True)
```

PR #63, opened 2020-08-28
Schema updates: SpectralTypes and Gravities

dr-rodriguez:
This adds the SpectralTypes and Gravities tables, as described in #21 and #22. 
It also adds tests to verify that any entries to those tables do include those values and that at least one best spectral type has been flagged.
Note that spectral types are encoded as 10-character strings as opposed to numeric floats (ie, we should expect things like L3.5, not just 13.5).

In addition, there are minor updates to the README and to the generate_database script in case users want to use the later to build/refresh a postgres version of the SIMPLE database.


scripts/tutorials/generate_database.py
```diff
@@ -7,14 +7,34 @@
...
+# SQLlite settings
+DB_TYPE = 'sqlite'
+DB_NAME = 'SIMPLE.db'
+
+# Set correct connection string
+if DB_TYPE == 'sqlite':
+    # First, remove the existing database in order to recreate it from the schema
+    # If the schema has not changed, this part can be skipped
+    if os.path.exists(DB_NAME):
+        os.remove(DB_NAME)
```
Will-Cooper, line 24: Is there a handle for if a user doesn't have write/ remove permissions?

simple/schema.py
```diff
@@ -32,6 +33,25 @@ class Instruments(Base):
...
+    radio = 'radio'
+
+
+class Gravity(enum.Enum):
+    """Enumeration for gravity"""
+    a = 'alpha'
+    b = 'beta'
+    g = 'gamma'
+    bg = 'beta/gamma'
+    unknown = 'unknown'
```
Will-Cooper, line 53: What about the ones from NIR spectra? e.g. vl-g, int-g, fld-g.
There's also delta for optical

scripts/tutorials/generate_database.py
dr-rodriguez: Nope, it's meant for you to run on your own machine or by the github action which already has permissions.

simple/schema.py
dr-rodriguez: Yep, good points- I don't know all the possibilities so it's good to get feedback like this!

dr-rodriguez:
I'll add the other gravity enumerations and change `best` to `adopted` in the table

tests/test_integrity.py
```diff
@@ -259,6 +259,44 @@ def test_radialvelocities(db):
...
+    if len(t) > 0:
+        print("\nSpectral Type entries with incorrect 'best' labels")
+        print(t)
+    assert len(t) == 0
+
+
+def test_gravities(db):
+    # Tests against the Gravities table
+
+    # There should be no entries in the SpectralTypes table without a spectral type
```
dr-rodriguez, line 290: Comment needs updating: Gravities

Will-Cooper (approved):

PR #69, opened 2020-09-15
Adding Spectra Table

dr-rodriguez:
Adding Spectra table, along with an example demonstrating how it can be used.
Requires latest version of Astrodbkit2 to be installed from the repo


dr-rodriguez:
Add observation_date as a primary key
Consider revisiting Modes to have instrument and telescope as primary keys- in case we want to add things like resolution in the future. Not sure how that would work with our foreign key setup.

data/Modes.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "name": "Prism",
+        "instrument": "SpeX",
```
Will-Cooper, line 4: Does case matter/ break anything? e.g. "spex" == "SpeX"?

scripts/tutorials/single_spectra_example.py
```diff
@@ -0,0 +1,42 @@
...
+              'instrument': 'SpeX',
+              'mode': 'Prism',
+              'reference': 'Cruz18',
+              'wavelength_units': 'um',
+              'flux_units': 'erg s-1 cm-2 A-1',
+              'observation_date': datetime.fromisoformat('2004-11-08')
+              }]
+db.Spectra.insert().execute(spec_data)
+
+_ = db.inventory('2MASS J00192626+4614078', pretty_print=True)
```
Will-Cooper, line 30: perhaps better variable names than the unnamed default "_"? (also, I can't see this being used?)

simple/schema.py
```diff
@@ -147,7 +154,38 @@ class Gravities(Base):
     __tablename__ = 'Gravities'
     source = Column(String(100), ForeignKey('Sources.source', ondelete='cascade', onupdate='cascade'),
                     nullable=False, primary_key=True)
-    gravity = Column(Enum(Gravity))  # restricts to enumerated values
+    gravity = Column(Enum(Gravity), nullable=False)  # restricts to enumerated values
```
Will-Cooper, line 157: why do we not want any of the above to be null? e.g. you can have sources with parallaxes and no proper motion in the same entry, no?

Will-Cooper:
Also the conflicting file issue with ```SIMPLE.db```?

scripts/tutorials/single_spectra_example.py
dr-rodriguez: The unnamed variable _ is a throwaway. This is just because db.inventory returns output that I don't want to see.

simple/schema.py
dr-rodriguez: Those are in different tables. This is saying that if you have an entry in the **Gravities** table, it should have a **gravity** value, otherwise you shouldn't be adding an entry here. It's OK for a source to have entries in one table, but not in another (such as parallaxes but no gravity measurements).

dr-rodriguez:
> Also the conflicting file issue with `SIMPLE.db`?

Yea, we'll have to ignore the binary file differences. This is why we use the python + json files to track database changes, there is no way for us to know what has changed in the binary file without explicitly loading it up.

data/Modes.json
dr-rodriguez: I believe this is case insensitive, but worth double checking.

kelle:
I think we should hold off in this PR until we're actually ready to start ingesting spectra. There are many discussions which need to be had about how to best format this table.

dr-rodriguez:
Withdrawn for now.

PR #85, opened 2020-11-03
Documentation: First Pass

dr-rodriguez:
This is a first pass at the documentation, focusing on the _Sources_ table.

As an aside, did I get **equinox** and **epoch** backwards? This is how it is in the schema, but that could be a mistake.


dr-rodriguez:
I've added the _Names_ and _Photometry_ tables, but for Photometry I was again unsure what we meant by **epoch**. Is this the same as the **epoch** in _Sources_? The datatype is different and if we were planning on storing date/time, there is a more appropriate DateTime datatype for that.

Will-Cooper:
> I've added the _Names_ and _Photometry_ tables, but for Photometry I was again unsure what we meant by **epoch**. Is this the same as the **epoch** in _Sources_? The datatype is different and if we were planning on storing date/time, there is a more appropriate DateTime datatype for that.

Think we can just store them always as floats, should be the same as in _Sources_, e.g. Gaia DR2 has epoch 2015.5 but 2MASS may be 1998 or whatever, but the equinox remains as J2000 (string)? Need to have exact epochs in _Photometry_ so people can look at variability.

dr-rodriguez:
I've created PR #86 to address the fact that equinox and epoch are backwards.
For handling variability, we should create an issue for that since the current schema may not be able to handle it, due to the primary keys in the _Photometry_ table.

documentation/Sources.md
```diff
@@ -0,0 +1,16 @@
...
+The Sources table contains all objects in the database alongside their coordinates. 
+This is considered the 'primary' table in the database, as each source 
+is expected to be unique and is referred to by all other object tables.
+
+| Column Name | Description  | Unit  | Data Type | Key Type  |
+|---|---|---|---|---|
+| source    | Unique identifier for the source |   | String(100)  | primary  |
+| ra        | Right Ascension | deg | Float  |   |
+| dec       | Declination | deg | Float  |   |
+| epoch     | Epoch for coordinates (eg, J2000) |   | String(10) |   |
```
kelle: Needs to be updated.

kelle (approved):

Issue #83, opened 2020-11-03
make documentation for each table in schema


kelle:
We need a place to keep track of the details for each table in the database. In particular, the rationale behind the Adopted column or more complex/non-typical fields. We decided to use the documentation folder and to make a .md file for each table. @dr-rodriguez will get it started.

kelle:
Documentation Needed

# Sources
- [ ] Document our standards for Epoch, Equinox (see #86)

# Photometry
- [ ] <>

# Astrometry
- [ ] <>

# More Tables
- [ ] <>

PR #86, opened 2020-11-12
Fixing equinox and epoch in database

dr-rodriguez:
The implementation of equinox and epoch in the database were backwards. I had equinox as the decimal year and epoch as the string designation of the equinox (ie, B1950, J2000). This swaps and also sets epoch in the _Photometry_ table to have the same Float type as in the _Sources_ table. 
Fortunately, we have no data yet for either of these so no updates needed for the individual JSON output.


kelle:
We should document our standard for this (see #83 ), but merging this in the meantime.

kelle (approved):
LGTM

PR #89, opened 2020-11-24
Add new environmnent file and update readme and getting started. Fixes #88

### LINKED ISSUE #88: Update environment.yml
Current enviornment.yml has lots of extra stuff.

kelle:


README.md
```diff
@@ -1,13 +1,27 @@
 # SIMPLE
 
-The vision of the SIMPLE project is to create a *collaborative* database of low-mass stars, brown dwarfs, and directly imaged exoplanets: a simple archive of complex objects. We plan to include coordinates, photometry, kinematics, spectra, images. We are also considering ways to include modeled and retrieved parameters. We are developing methods to interact with the database via several methods, including python, a website and API, and database browsers.
+The vision of the SIMPLE project is to create a *collaborative* database of low-mass stars, brown dwarfs, and directly 
+imaged exoplanets: a simple archive of complex objects. The tables and fields currently included in the 
+database are described in the [Documentation](https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/documentation/README.md)
```
dr-rodriguez, line 5: I would suggest we use relative paths (see, for example, the way the image is done) rather than absolute http links. That also allows for different documentation across branches, in case that situation comes up.

dr-rodriguez:
Why did you create a new environment file instead of replacing the existing one? With this merge we would have environment.yml and environment2.yml...

README.md
```diff
@@ -16,50 +30,72 @@ If you'd like to set up your own copy of the SIMPLE database, here's what we rec
...
-```
-
-7. Load database contents from source, if needed. 
+A conda environment file exists for convenience. The following commands will create and activate an environment called `simple-db`:
+
+    ```bash
+    conda env create -f environment2.yml
+    conda activate simple-db
+    ```
+    <!---
```
dr-rodriguez, line 39: Why not delete all this commented code? It will still be in the version history should we need it.

environment2.yml
```diff
@@ -0,0 +1,51 @@
...
+  - setuptools=50.3.1=py39hecd8cb5_1
+  - sqlite=3.33.0=hffcf06c_0
+  - tk=8.6.10=hb0a8c7a_0
+  - tzdata=2020d=h14c3975_0
+  - wheel=0.35.1=pyhd3eb1b0_0
+  - xz=5.2.5=h1de35cc_0
+  - zlib=1.2.11=h1de35cc_3
+  - pip:
+    - asdf==2.7.1
+    - astrodbkit2==0.2.1.dev7+g1909568
```
dr-rodriguez, line 24: I think this will fail since nothing here specifies that pip should use the github version of this package. 
In the long run (and for stability), I think it's OK that users just do pip install astrodbkit2 and not go the github route. 
We can do that since we're still in active development.

kelle: so should I delete this from the environment file?

kelle:
I made envorinment2.yml because I didn't want to delete the file @Will-Cooper made. Will, could you take a look at this new file and see if this new one does the same thing as you want it to? If so, I can overwrite the old one.

Will-Cooper:
That environment2.yml looks good to me @kelle! Sorry, been super busy recently...  
Up to you whether you want the pip version of astrodbkit2 or the github version seeing as both simple & astrodbkit2 are still in active development. (We can always update to the stable pip version when we approach a public release)

kelle:
This is ready for re-review. :)

Will-Cooper:
I see astrodbkit2 isn't in the environment.yml, is this because we're telling users to manually install using pip (rather than just pip install in the environment.yml, which would also do it)?

kelle:
Yes, I removed astrodbkit2 from the environment file because we are explicitly telling folks to install from github.

Will-Cooper (approved):

dr-rodriguez (approved):
This all looks OK to me.

Issue #82, opened 2020-11-03
Ingest Y dwarf data


kelle:
Ingest Y dwarf sources and other information. It will be a good test of the new `add_table_data`. 

- [x] sources and coords
- [x] parallaxes
- [x] proper motions
- [ ] photometry
- [x] spectral type
- [ ] test new content

dr-rodriguez:
@cfontanive we had to do a bit of git magic, but we were able to push your Sources and Names changes via my branch. You will want to sync up with the latest changes in upstream as you continue working on other data you want to add.

kelle:
now that the sources are ingested, I think we decided to work on spectral types next, right?

cfontanive:
Yes, that's what I'm doing next!

kelle:
Start thinking about what data content tests we could add for this dataset.

PR #91, opened 2020-12-08
Adding Y dwarf sources; #82

cfontanive:


scripts/ingests/Y_dwarf_source_ingest.py
```diff
@@ -0,0 +1,42 @@
...
+existing_sources = []
+missing_sources = []
+for i,name in enumerate(Ydwarfs['source']):
+	if len(db.search_object(name,resolve_simbad=True)) != 0:
+		existing_sources.append(i)
+	else:
+		missing_sources.append(i)
+
+# add missing references
+temp = db.query(db.Publications.c.name).all()
```
dr-rodriguez, line 27: As mentioned during the meeting, here is a more compact way of doing most of lines 27-33:
```
ref_list = Ydwarfs['reference'].tolist()
included_ref = db.query(db.Publications.c.name).filter(db.Publications.c.name.in_(ref_list)).all()
included_ref = [s[0] for s in included_ref]
new_ref = list(set(ref_list)-set(included_ref))
```
The first line gets a list of the sources from your csv file.
The second line gets all database entries that match any in your list
The third line is just some quick formatting of the result list
The fourth line takes the difference: it outputs all references from ref_list that are not among those returned by the database query.

dr-rodriguez: Actually, I guess an extra step would be needed to convert it to a dictionary:
`new_ref = [{'name': s} for s in new_ref]`

kelle (approved):

dr-rodriguez (approved):

Issue #90, opened 2020-12-01
Ingest Cloud Atlas dataset of HST data for brown dwarfs


kelle:
ref: https://arxiv.org/abs/1812.03963

Table 1 has Name, SpT, RA, Dec, 2MASS JHK, parallax (+ref). Could also do spectra. 

No discovery reference and no reference for spectral type. However, I suspect majority are already in the database via BDNYC and #48. 

Also going to use new astrodbkit method `add_table_data` to ingest data via a CSV file (#79 )

- [x] sources, coords, names
- [ ] ~~photometry~~
- [ ] parallaxes
- [x] spectral types [#130]
- [ ] spectra

kelle:
Spectral types ingested in #130

kelle:
Vizier link to the table: http://vizier.cfa.harvard.edu/viz-bin/VizieR-3?-source=J/AJ/157/101/table1

PR #106, opened 2021-02-09
Ingesting ATLAS sources, #90

NiallWhiteford:


kelle:
I think the next step is to write a script which uses a RA-dec search to find a simbad match if the simbad match fails. And then present the user with the option to select the nearest object. This should automate the matching for most 2MASS and WISE sources with good coords but improperly formatted designations.

scripts/ingests/ingest_ATLAS_will_work.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+
+
+# find sources already in database
+existing_sources = []
+missing_sources = []
+db_names = []
+for i,name in enumerate(ATLAS['Name']):
+	if len(db.search_object(name,resolve_simbad=False)) != 0:
```
dr-rodriguez, line 68: We should consider having a flag variable, `simbad_check`, which should always be set to `True` when running this loop. For cases we want to skip the checking (like S Ori 70), we can temporarily set `simbad_check=False`. This will allow us to skip Simbad name verification for cases we know won't work.

dr-rodriguez:
I quickly pushed a change to this PR to remove the duplicated entries. I did this by running db.save_db again. I still need to debug why they were duplicated in the first place.

kelle (approved):

PR #123, opened 2021-04-06
update utils

zjzhang42:
- `verbosprint` the object name if no SIMBAD counterparts are found
- make `radius` as a keyword in the function `check_names_simbad `
- make sure `coord_result_table` is not `None` before getting its length.


kelle (approved):
Looks great!!

dr-rodriguez (approved):

PR #131, opened 2021-04-29
Auto-generate database

dr-rodriguez:
This update re-enables the Github Action to generate the database for closed pull requests. The binary file is then pushed to a new repo just for it: https://github.com/SIMPLE-AstroDB/SIMPLE-binary 
This can also be run manually from the Actions page and we can update it to set it on a schedule.

I've also added a .gitignore file that lists the binary database file so we can avoid pushing changes to it. At some point later we will want to explicitly remove it from the repo entirely.

### DIFF
--- a/.github/workflows/gen-db.yml
+++ b/.github/workflows/gen-db.yml
@@ -5,7 +5,11 @@
 name: Generate Database
 
 on:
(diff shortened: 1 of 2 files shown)


kelle (approved):
LGTM

PR #142, opened 2021-05-11
Add utils functions to search and add publications. Fixes #137

### LINKED ISSUE #137: Make a utils function for adding a reference
Function could complicated and do the DOI resolving for the user.

```
# Adding new reference Manj19 to publications table in database
manj19_search = db.query(db.Publications).filter(db.Publications.c.name == 'Manj19').table()
if len(manj19_search) == 0 and not DRY_RUN:
	new_ref = [{'name': 'Manj19'}]
	# Should have included bibcode and doi
	# new_ref = [{'name': 'Manj19', 'bibcode': '2019AJ....157..101M', 'doi': '10.3847/1538-3881/aaf88f'}]
	db.Publications.insert().execute(new_ref)
```

kelle:


Will-Cooper (commented):
```search_publications``` looks good to me, we tried multiple edge cases (lower case names, A&A bibcodes, no kwargs).

**Comment by kelle** [comment]
Even though there are still some bells and whistles which could be added to the add_publication script, I think it's good to merge now.  It works very well when the doi is provided. :)

File: scripts/ingests/utils.py [hunk 1]
```diff
@@ -1,10 +1,253 @@
...
+    --------
+    search_publication: Function to find publications in the database
+
+    """
+
+    if not(doi or bibcode):
+       print('DOI or Bibcode is required input')
+       return
+
+    ads.config.token = os.getenv('ADS_TOKEN')
```
dr-rodriguez, line 141: Should we error out if ADS_TOKEN is not set?

scripts/ingests/utils.py
```diff
@@ -1,10 +1,253 @@
...
+        Name of publication to search
+    doi: str
+        DOI of publication to search
+    bibcode: str
+        ADS Bibcode of publication to search
+
+    Returns
+    -------
+    True: if only one match
+    Table containing publications matching name if more than one match
```
dr-rodriguez, line 32: I worry that `search_publication` returning different types of output can cause issues further along the road. I would prefer if it either always returns a table (even if the table is empty) or always returns a boolean True/False, not the mix of the two.

kelle: I agree! I was hoping you would tell me to do this. :)

scripts/ingests/utils.py
```diff
@@ -1,10 +1,253 @@
...
+            print('Use add_publication() to add it to the database.')
+            return False
+
+        if n_pubs_found_short > 0:
+            print(f'Found {n_pubs_found_short} matching publications for {shorter_name}')
+            if verbose:
+                pub_search_table.pprint_all()
+            return pub_search_table
+
+    return
```
kelle, line 108: Any thoughts on this final empty `return`?  I kinda just put it there to "close" the function but all the cases should have been covered above. @blackwer?

scripts/ingests/utils.py
kelle: No, I want the function to still work even without ADS. I'll add more feedback though.

scripts/ingests/utils.py
```diff
@@ -1,10 +1,253 @@
...
+
+        if len(bibcode_matches_list) == 1:
+            print("Publication found in ADS using bibcode: ", bibcode)
+            article = bibcode_matches_list[0]
+            print(article.first_author, article.year, article.bibcode, article.title)
+            if not name:  # generate the name if it was not provided
+                name = article.first_author[0:4] + article.year[-2:]
+            description = article.title[0]
+
+    #check again to make sure publication does not already exist in database
```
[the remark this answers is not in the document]
kelle: Brilliant.

kelle:
This is ready for re-review. @dr-rodriguez

tests/test_utils.py
```diff
@@ -79,3 +79,14 @@ def test_ingest_parallaxes(db, t):
     assert results['source'][0] == 'Fake 3'
     assert results['parallax'][0] == 155
     assert results['parallax_error'][0] == 0.6
+
+def test_add_publication(db):
```
dr-rodriguez, line 83: This is breaking the tests. Either complete the test, or add an explicit `return` to continue

dr-rodriguez (approved):

dr-rodriguez:
Looks good. We can probably tweak the tests more, but for now we can merge.

PR #148, opened 2021-06-21
Y Dwarf Photometry Ingest Script

ebhort27:
Started working on script for ingesting the Y Dwarf photometry data


scripts/ingests/Y_dwarf_photometry_ingest.py
```diff
@@ -0,0 +1,67 @@
...
+from astrodbkit2.astrodb import Database
+from simple.schema import *
+from utils import ingest_photometry
+from astropy.table import Table
+from pathlib import Path
+import os 
+
+
+
+#From ingest_ATLAS_spectral_types.py file
```
kelle, line 14: Remove this comment.
```suggestion
```

scripts/ingests/utils.py
```diff
@@ -204,3 +205,35 @@ def ingest_pm(db, sources, muRA, muRA_err, muDEC, muDEC_err, pm_reference, verbo
         n_added += 1
 
     print("Added to database: ", n_added)
+#def ingest_photometry adapted from ingest_pm and ingest_parallaxes from utils
```
kelle, line 208: Delete this comment
```suggestion
```

scripts/ingests/utils.py
```diff
@@ -204,3 +205,35 @@ def ingest_pm(db, sources, muRA, muRA_err, muDEC, muDEC_err, pm_reference, verbo
...
+        # Construct data to be added
+        photometry_data = [{'source': db_name,
+                          'band': bands[i],
+                          'ucd' : ucds[i],
+                          'magnitude' : magnitudes[i],
+                          'magnitude_error' : magnitude_errors[i],
+                          'telescope': telescope[i],
+                          'instrument': instrument[i],
+                          'epoch': epoch,
+                          'comments': comments,
```
kelle, line 230: This works when `epoch` and `comments` are None but will not have the expected behavior if passed as an array. Try adding the indexes.

scripts/ingests/utils.py
```diff
@@ -204,3 +205,35 @@ def ingest_pm(db, sources, muRA, muRA_err, muDEC, muDEC_err, pm_reference, verbo
         n_added += 1
 
     print("Added to database: ", n_added)
+#def ingest_photometry adapted from ingest_pm and ingest_parallaxes from utils 
+def ingest_photometry(db, sources = None, bands = None, ucds = None, magnitudes = None, magnitude_errors = None , telescope = None, instrument = None, epoch = None, comments = None, reference = None, verbose=False):
+    verboseprint = print if verbose else lambda *a, **k: None
+
```
kelle, line 211: you might need to do something like this for every optional array (ucd, telescope, instrument, comment, epoch)
```suggestion
if epoch = None:
        epoch = None*len(sources)
```

kelle (changes requested):

scripts/ingests/utils.py
ebhort27: So I originally had the indexes in there, but it kept tracebacking because it says that Nonetype object wasn't subscriptable, so I took them out for now

tests/test_data.py
```diff
@@ -136,4 +136,8 @@ def test_Kirk19_ingest(db):
     ref = 'Kirk19'
     t = db.query(db.ProperMotions).filter(db.ProperMotions.c.reference == ref).astropy()
     assert len(t) == 22, f'found {len(t)} proper motion entries for {ref}'
-    
+
+    #Test photometry added
```
kelle, line 140: Also add a test for photometry with the "Kirk19" reference.

kelle (changes requested):
This looks GREAT!  Please add one more test and then I think it'll be good to merge. :)

tests/test_data.py
ebhort27: So the only thing I'm a bit confused about is that it looks like there are 9 objects in the csv file with Kirk19 as the spitzer reference, but 16 total in the database now (which means 8 objects, since each is in there twice). Did it miss one? Should I put "assert len(t) == 16" or assert len(t) == 18"?

kelle: Maybe one of the Kirk19 sources is the first one which we are skipping? 

And i forgot about the other ref...add that as a test too. :)

ebhort27: It's the first one which we're skipping! That explains it. I'll add a test for the other source too

kelle (approved):

PR #162, opened 2021-07-01
2020 Best Source Ingest Script

ebhort27:
This will be the ingest script for the 2020 Best Sample sources

### DIFF
--- a/scripts/ingests/2020Best_publications_ingest.py
+++ b/scripts/ingests/2020Best_publications_ingest.py
@@ -12,7 +12,7 @@ from pathlib import Path
 import os
 
 save_db = False #modifies .db file but not the data files
-RECREATE_DB = False #recreates the .db file from the data files
+RECREATE_DB = True #recreates the .db file from the data files
 VERBOSE = False
 
 verboseprint = print if VERBOSE else lambda *a, **k: None
@@ -34,13 +34,13 @@ else:
 
 # add missing references
 #Searching for all publications in table and adding missing ones to pub table in .db file
-Pubs = Table.read('scripts/ingests/UltracoolSheet-References.csv', data_start=1, data_end=20)
+Pubs = Table.read('scripts/ingests/UltracoolSheet-References.csv', data_start=1, data_end=50)
 bibcodes = Pubs['ADSkey_ref']
 for i,ref in enumerate(Pubs['code_ref']):
-	if search_publication(db, name=ref) == False:
-		if search_publication(db, bibcode=bibcodes[i]) == False:
-			add_publication(db, name=ref, bibcode=bibcodes[i], save_db=save_db)
-		elif search_publication(db, bibcode=bibcodes[i], verbose=True) == True:
+	if search_publication(db, bibcode=bibcodes[i]) == False:
+		add_publication(db, name=ref, bibcode=bibcodes[i], save_db=save_db)
+	elif search_publication(db, bibcode=bibcodes[i], verbose=True) == True:
+		if search_publication(db, name=ref) == False:
 			print('Mismatched Source:',ref)
 	elif search_publication(db, name=ref) == True:
 		pass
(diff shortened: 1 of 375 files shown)


scripts/ingests/2020Best_source_ingest.py
```diff
@@ -0,0 +1,78 @@
...
+
+# add missing references
+ref_list = Ydwarfs['reference'].tolist()
+included_ref = db.query(db.Publications.c.name).filter(db.Publications.c.name.in_(ref_list)).all()
+included_ref = [s[0] for s in included_ref]
+new_ref = list(set(ref_list)-set(included_ref))
+new_ref = [{'name': s} for s in new_ref]
+
+if len(new_ref)>0:
+	db.Publications.insert().execute(new_ref)
```
kelle, line 55: Get rid of this...this is now taken care of by `search_publication` and `add_publication` functions.

scripts/ingests/2020Best_source_ingest.py
```diff
@@ -0,0 +1,78 @@
...
+# add missing objects to Sources table
+if len(missing_sources)>0:
+	db.add_table_data(Ydwarfs[missing_sources], table='Sources', fmt='astropy')
+
+# add new sources in Names table too
+names_data = []
+for ms in missing_sources:
+	names_data.append({'source': Ydwarfs['source'][ms], 'other_name':Ydwarfs['source'][ms]})
+if len(missing_sources)>0:
+	db.Names.insert().execute(names_data)
```
kelle, line 67: make this a stand alone function. Can stay here while you write it, but will eventually move to utils.

scripts/ingests/2020Best_source_ingest.py
```diff
@@ -0,0 +1,78 @@
...
+sys.path.append('.')
+from astrodbkit2.astrodb import create_database
+from astrodbkit2.astrodb import Database
+from simple.schema import *
+from utils import *
+from astropy.table import Table
+from pathlib import Path
+import os
+
+DRY_RUN = False #modifies .db file but not the data files
```
kelle, line 12: change `dry_run` to `save_db`

scripts/ingests/2020Best_source_ingest.py
```diff
@@ -0,0 +1,78 @@
...
+# add other names for existing sources if alternative names not in database yet
+other_names_data = []
+for es in existing_sources:
+	es_names = db.search_object(db_names[es], output_table='Names')
+	if Ydwarfs['source'][es] not in [x[1] for x in es_names]:
+		other_names_data.append({'source': db_names[es], 'other_name':Ydwarfs['source'][es]})
+if len(existing_sources)>0:
+	db.Names.insert().execute(other_names_data)
+
+db.save_db('../../data')
```
kelle, line 78:
```
if save_db == True:

File: scripts/ingests/2020Best_source_ingest.py (hunk 1 above)
Comment (ebhort27) [reply scripts/ingests/2020Best_source_ingest.py:55]: So should I instead set up a function where it uses search_publication to find the missing publications and then use add_publication to add them?

Comment (kelle) [reply scripts/ingests/2020Best_source_ingest.py:55]: doesn't even have to be another function.
```
if search_publication(ref) == False:
    add_publication(ref)

ebhort27: The issue is, when I used the add_publication function, it says it needs a name and bibcode to be entered

scripts/ingests/2020Best_publications_ingest.py
```diff
@@ -0,0 +1,71 @@
...
+else:
+    db = Database(db_connection_string) #if database already exists, connects to .db file
+
+
+# add missing references
+#Searching for all publications in table and adding missing ones to pub table in .db file
+Pubs = Table.read('scripts/ingests/UltracoolSheet-References.csv', data_start=1, data_end=20)
+bibcodes = Pubs['ADSkey_ref']
+for i,ref in enumerate(Pubs['code_ref']):
+	if search_publication(db, name=ref) == False:
```
kelle, line 40: I think we should get rid of this first `if`.  Just keep the next one. In other words, if the bibcode doesn't exist in the database, add it.

PR #167, opened 2021-07-13
BEST 2020 Sample Publications Ingest

ebhort27:
This is a new pull request just for the scripts related to ingesting the publications from the BEST 2020 Sample. There will later be a separate pull request for the source ingest.

### DIFF
--- a/data/Publications.json
+++ b/data/Publications.json
@@ -42,12 +42,6 @@
         "description": "DENIS J081730.0-615520: An Overlooked Mid-T Dwarf in the Solar Neighborhood"
     },
     {
-        "name": "Barm10",
-        "bibcode": "2011ApJ...733...65B",
-        "doi": "10.1088/0004-637X/733/1/65",
-        "description": "Clouds and Chemistry in the Atmosphere of Extrasolar Planet HR8799b"
-    },
-    {
         "name": "Bene99",
         "bibcode": "1999AJ....118.1086B",
         "doi": "10.1086/300975",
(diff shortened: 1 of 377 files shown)


kelle:
This looks amazing!!!  The only problem is with file data/wisea_j114156.67-332635.5.json where a spectral type measurement is deleted.  Apparently, deleting the `Legg17` publication cascading up to delete all related records. I think we need to do some git magic to revert that one file. Let's discuss on slack.

data/wisea_j114156.67-332635.5.json
```diff
@@ -58,16 +58,5 @@
             "comments": null,
             "reference": "Kirk19"
         }
-    ],
-    "SpectralTypes": [
```
kelle: We need to not delete this measurement.

scripts/ingests/2020Best_publications_ingest.py
```diff
@@ -0,0 +1,180 @@
...
+add_Irwi91_bibcode = db.Publications.update().where(db.Publications.c.name == 'Irwi91').\
+	values(bibcode='1991MNRAS.252P..61I')
+db.engine.execute(add_Irwi91_bibcode)
+update_Kirk97 = db.Publications.update().where(db.Publications.c.name == 'Kirk97b').\
+	values(name='Kirk97a')
+db.engine.execute(update_Kirk97)
+update_Law06 = db.Publications.update().where(db.Publications.c.name == 'Law_06').\
+	values(name='Law_06b')
+db.engine.execute(update_Law06)
+db.Publications.delete().where(db.Publications.c.name == 'Legg17').execute()
```
kelle, line 64: This is the action which caused the problem. I don't think we should change the script at this point, but just a lesson learned.

data/Publications.json
```diff
@@ -3730,5 +3706,2795 @@
         "bibcode": "2016AJ....152...78L",
         "doi": null,
         "description": null
+    },
+    {
+        "name": "Legg17",
+        "bibcode": "2017ApJ...842..118L",
+        "doi": "10.3847/1538-4357/aa6fb5",
+        "description": "The Y-type Brown Dwarfs: Estimates of Mass and Age from New Astrometry, Homogenized Photometry, and Near-infrared Spectroscopy"
+    },
```
kelle, line 3715: `Legg17` was added correctly to the Publications table, so we actually don't need to modify this table anymore!!  Huzzah!

kelle (approved):

Issue #105, opened 2021-01-26
Ingestion of sources not in Simbad


cfontanive:
How do we check whether a source already exists in the database if it is not in Simbad? (e.g. WISEA J030237.52-581740.3 in the Y dwarf table)

kelle:
It looks like you took care of this issue in your [Y_dwarf_source_ingest.py](https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/scripts/ingests/Y_dwarf_source_ingest.py) script. We'll eventually get to using that method in the ingesting ATLAS script as well.

Do we need to keep this issue open or was there something else we wanted to do here?

cfontanive:
No, I am currently ignoring this source in my script, which is why I started this issue. I currently use Simbad to check whether a source is already in the database or not, but we need a way to do this for cases of sources that do not exist in Simbad (not just under a different name).

kelle:
Ah, i see, you're ignoring it with `data_start=2` on line 16? 

So we need to incorporate the Simbad search thing we've added to `ingest_ATLAS_will_work.py`, currently a WIP in #106 . I think you can copy and paste lines 31-63 and then loop over `resolved_name` instead of `Ydwarfs['source']`. Sound like a plan?

kelle:
nope, nope, not a plan at all. We need to fill out the final `else` in that script (line 62-63). (And probably turn it into a function!!). I think we need to make a 2nd empty list which gets populated. Maybe we make a  `not_in_simbad` list to partner with `resolved_name`. So if both the identifier and coord searches failed, the `ATLAS['Name']` gets appended to `not_in_simbad`.

cfontanive:
Yes that source is the first one in my list, so I'm currently skipping it.
This won't help – that Simbad loop was made to identify the Simbad names of sources that were not recognised in Simbad with the names from the ATLAS list. The source listed above does not exist in Simbad at all, so we currently have no way to check whether it is already in the database.

cfontanive:
> nope, nope, not a plan at all. We need to fill out the final `else` in that script (line 62-63). (And probably turn it into a function!!). I think we need to make a 2nd empty list which gets populated. Maybe we make a `not_in_simbad` list to partner with `resolved_name`. So if both the identifier and coord searches failed, the `ATLAS['Name']` gets appended to `not_in_simbad`.

[...]

PR #169, opened 2021-07-15
More utils functions for ingesting sources.

kelle:
Adds `sort_sources` function and ability to search by location as discussed in #105. 

Also `ingest_sources` and `add_names` functions.

Tests not written yet but will do at some point!


ebhort27 (approved):
Looks good!

PR #174, opened 2021-07-26
Best 2020a Proper Motion Ingest

ebhort27:
This will be the script for the Best 2020 Sample proper motion ingest for the ones with reference Best20a


kelle (approved):
Looks great! Great job!

PR #175, opened 2021-07-28
Best 2020 Proper Motion Lit Ingest

ebhort27:
This will be the PR for the 2020 Best proper motion lit ingest.


kelle:
This is looking good!  Regarding alt names, ones we should look into:
```
'2MASS J12560183-1257276', 'other_name': 'VHS J125601.92-125723.9 b'})
'2MASS J12560183-1257276', 'other_name': 'VHS J125601.92-125723.9 AB'})
```
and
```
'2MASS J15470557-1626303', 'other_name': 'PSO J236.7729-16.4422'})
'2MASS J15470557-1626303', 'other_name': '2MASS J15470557-1626303B'})
```
I think these might be different sources being matched to the same source in the database...just need to double check.

kelle:
- VHS J125601.92-125723.9 AB is a new source. Lets use 2MASS J12560215-1257217 as it's primary name and VHS as an Other Name.
- 2MASS J15470557-1626303B should be a new source.

kelle:
I think this script is ready to run once the Proper Motion table gets an `adopted` column (#180 ).

kelle:
Also need to add some tests!

dr-rodriguez (approved):

PR #177, opened 2021-08-03
Keep track of new other names in the `sort_sources` function.

kelle:
Now the `sort_sources` function keeps track of alternate identifiers which need to be added to the Names table. It outputs an Astropy table. And now the `add_names` table expects that table as an input.

### DIFF
--- a/scripts/ingests/utils.py
+++ b/scripts/ingests/utils.py
@@ -27,13 +27,38 @@ def disable_exception_traceback():
     sys.tracebacklimit = default_value  # revert changes
 
 
-def sort_sources(db, ingest_names, ingest_ras, ingest_decs, verbose=False, search_radius=60.):
+def sort_sources(db, ingest_names, ingest_ras, ingest_decs, search_radius=60., verbose=False):
+    '''
+
+    Parameters
+    ----------
+    db
+    ingest_names
+        Names of sources
+    ingest_ras
+        Right ascensions of sources. Decimal degrees.
+    ingest_decs
+        Declinations of sources. Decimal degrees.
+    search_radius
+        radius in arcseconds to use for source matching
+    verbose
+
+    Returns
+    -------
+    missing_sources_index
+        Indices of sources which are not in the database
+    existing_sources_index
+        Indices of sources which are already in the database
+    alt_names_table
+        Astropy table with Other Names to add to database
+
+    '''
 
     verboseprint = print if verbose else lambda *a, **k: None
 
     existing_sources_index = []
     missing_sources_index = []
-    alt_names_table = Table(names=('ingest_name','db_name'),dtype=('U30','U30'))
+    alt_names_table = Table(names=('db_name','other_name'),dtype=('U30','U30'))
     db_names = []
 
     for i, name in enumerate(ingest_names):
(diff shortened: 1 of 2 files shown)


scripts/ingests/utils.py
```diff
@@ -27,12 +28,39 @@ def disable_exception_traceback():
...
+    alt_names_table
+        Astropy table with Other Names to add to database
+
+    '''
 
     verboseprint = print if verbose else lambda *a, **k: None
 
     existing_sources_index = []
     missing_sources_index = []
+    Alt_names = namedtuple("Alt_names", "source other_name")
```
dr-rodriguez, line 62: Variables names should not be capitalized, though this is just a convention.

scripts/ingests/utils.py
```diff
@@ -45,6 +73,10 @@ def sort_sources(db, ingest_names, ingest_ras, ingest_decs, verbose=False, searc
             verboseprint(i, ": no name matches, trying simbad search")
             try:
                 namematches = db.search_object(name, resolve_simbad=True, verbose=verbose)
+                if len(namematches) == 1:
+                    simbad_match = namematches[0]['source']
+                    # Populate astropy table with ingest name and database name match
```
dr-rodriguez, line 78: Technically, alt_names_table is a list, not an astropy table.

scripts/ingests/utils.py
```diff
@@ -75,29 +110,66 @@ def sort_sources(db, ingest_names, ingest_ras, ingest_decs, verbose=False, searc
...
     print(n_missing, "sources not found in the database")
 
-    return missing_sources_index, existing_sources_index, db_names
+    return missing_sources_index, existing_sources_index, alt_names_table
 
 
-def add_names(db, new_sources, verbose=True, save_db=False):
+def add_names(db, sources=None, other_names=None, names_table=None, verbose=True, save_db=False):
+    '''
+    Provide either two lists of sources and other_names or a 2D name_table.
```
dr-rodriguez, line 136: Suggested rewrite:
```
Add source names to the Names table in the database. 
Provide either two lists of sources and other_names or a 2D names_table.
```

scripts/ingests/utils.py
```diff
@@ -56,6 +88,9 @@ def sort_sources(db, ingest_names, ingest_ras, ingest_decs, verbose=False, searc
             nearby_matches = db.query_region(location, radius=radius)
             if len(nearby_matches) == 1:
                 namematches = nearby_matches
+                coord_match = namematches[0]['source']
+                # Populate astropy table with ingest name and database name match
+                alt_names_table.append(Alt_names(coord_match, name))
             if len(nearby_matches) > 1:
                 print(nearby_matches)
                 raise Exception("too many nearby sources!")
```
dr-rodriguez, line 96: It's always better to raise a more specific exception (eg, TypeError, ValueError, RuntimeError) rather than the general Exception, but I don't know which one would be best.

scripts/ingests/utils.py
dr-rodriguez, line 62: Curious why the first entry is 'Alt_names' when it is meant to be the source name; the 'source other_name' is actually the alternate name.

dr-rodriguez: OK, clarified how this works so I'm fine with it.

dr-rodriguez (approved):
Looks good

Issue #163, opened 2021-07-06
Github Action to update binary repo


Will-Cooper:
Looks like this is still not automatically working, at least for @ebhort27; it doth however still work with a manual action?  
Related, in the long term this would be a good thing to trigger a rebuild of the website at the same time (it does right now anyway, but only because the website goes down after ~35mins of inactivity).

dr-rodriguez:
I've debugged this a little and the only thing I can see is a github redirect warning. I opened a ticket in the action repo to see if they have any more insight into this: https://github.com/dmnemec/copy_file_to_another_repo_action/issues/32

dr-rodriguez:
Just saw this in our Settings page: `Secrets are not passed to workflows that are triggered by a pull request from a fork.`

That means that the authorization token that allows it to write to SIMPLE-binary is not passed to the action and that may be why it's failing only on pull requests. This would be a limitation of github itself and we would have to figure out a way to work around it.

kelle:
close?

Will-Cooper:
Close if the next PR request triggers successfully I guess

PR #179, opened 2021-08-15
Updating database generation scripts

dr-rodriguez:
Main changes:

- Updating database generation script to produce either a sqlite binary file or load to a postgres database
- Update the Github-Actions script to generate both databases, loading the postgres one to a Heroku instance. This may allow us to completely avoid generating the binary file and dealing with pushing to a separate repo (ie, #163).

### DIFF
--- a/scripts/tutorials/generate_database.py
+++ b/scripts/tutorials/generate_database.py
@@ -1,6 +1,7 @@
 # Script to generate database from JSON contents
 # This gets run automatically with Github Actions
 
+import argparse
 import sys
 import os
 from astrodbkit2.astrodb import create_database, Database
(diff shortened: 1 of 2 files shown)


scripts/tutorials/generate_database.py
```diff
@@ -38,14 +31,48 @@
...
+                connection_string = sys.argv[2]
+            else:
+                connection_string = os.getenv('SIMPLE_DATABASE_URL')
+            load_postgres(connection_string)
+        elif architecture == 'sqlite':
+            load_sqlite()
+        else:
+            print(f'Unrecognized architecture: {architecture}')
+    else:
+        print('Specify database architecture (eg, postgres, sqlite)')
```
Will-Cooper, line 78: Not a huge fan of this implementation, you could do this with ```argparse```, no?

Will-Cooper (changes requested):
LGTM, will be curious to see if this fixes #163

Will-Cooper:
Also let me know if we'll also need to update SIMPLE-web to match with the ```wget```

scripts/tutorials/generate_database.py
dr-rodriguez: This was quick and dirty, but looking through `argparse` I like how it works so I'll update it.

Will-Cooper (approved):
LGTM, I like the ```argparse``` change for sure

PR #183, opened 2021-08-20
Ingest Gaia data script

kelle:


kelle:
This script successfully works on the .db file to modify records for 1266 sources:
```
Names added to database:  1266
Proper motions added to database:  1076
Parallaxes added to database:  1076
Grp: Photometry measurements added to database:  1106
G: Photometry measurements added to database:  1266
```
Once this PR is approved and merged, I'll do a separate PR modifying the 1266 JSON files.

scripts/ingests/utils.py
```diff
@@ -48,6 +52,26 @@ def verboseprint(*args, **kwargs):
 #     sys.tracebacklimit = default_value  # revert changes
 
 
+def load_simpledb(RECREATE_DB=True):
+    # Utility function to load the database
+
+    db_file = 'SIMPLE.db'
```
dr-rodriguez, line 58: I feel like `db_file` should be a parameter that the user can pass to the `load_simpledb` function, even it has a default value of `'SIMPLE.db'`

scripts/ingests/utils.py
```diff
@@ -721,26 +741,57 @@ def ingest_parallaxes(db, sources, plxs, plx_errs, plx_refs, save_db=False, verb
...
+    for i, source in enumerate(sources): # loop through sources with parallax data to ingest
         db_name = db.search_object(source, output_table='Sources')[0]['source']
 
-        # Search for existing parallax data.
+        # Search for existing parallax data and determine if this is the best
         # If no previous measurement exists, set the new one to the Adopted measurement
-        # TODO: figure out adopted in cases of multiple measurements.
         adopted = None
+        duplicate = False
         source_plx_data = db.query(db.Parallaxes).filter(db.Parallaxes.c.source == db_name).table()
```
dr-rodriguez, line 751: You can do something like: `db.search_object(source, output_table='Parallaxes')` to return the Parallaxes table for the matched source without having save a `db_name` variable that you use in the subsequent query. You still have access to the source name via the _source_ column in Parallaxes.

dr-rodriguez: Actually, I guess you may need to keep `db_name` for ingesting brand new parallaxes so maybe this isn't as useful.

scripts/ingests/utils.py
```diff
@@ -721,26 +741,57 @@ def ingest_parallaxes(db, sources, plxs, plx_errs, plx_refs, save_db=False, verb
...
             adopted = True
+            duplicate = False
+            old_adopted = None
+        elif len(source_plx_data) > 0:  # plx data already exists
+            for plx_data in source_plx_data:
+                if plx_data['reference'] == plx_refs[i]:  # check to see if other measurement is a duplicate of the new data
+                    duplicate = True
+                    verboseprint("Duplicate measurement\n", plx_data)
+                if plx_data['adopted'] == 1:  # check if another measurement is adopted
+                    old_adopted = plx_data
```
dr-rodriguez, line 764: You should not need to loop over each result to match references/adopted. Something like `ind = plx_data['reference'] == plx_refs[i]` should give you an array of True/False that you can sum up to see if any matches were found. Similarly for the checking of adopted check, which you can then use with `old_adopted = plx_data[ind]` to store that matched row.

scripts/ingests/utils.py
```diff
@@ -721,26 +741,57 @@ def ingest_parallaxes(db, sources, plxs, plx_errs, plx_refs, save_db=False, verb
...
+                if plx_errs[i] < min(source_plx_data['parallax_error']):
+                    adopted = True
+                    # unset old adopted
+                    if old_adopted:
+                        db.Parallaxes.update().where(and_(db.Parallaxes.c.source == old_adopted['source'],
+                                                             db.Parallaxes.c.reference == old_adopted['reference'])).\
+                            values(adopted=False).execute()
+                        old_adopted_data = db.query(db.Parallaxes).filter(and_(db.Parallaxes.c.source == old_adopted['source'],
+                                                                          db.Parallaxes.c.reference == old_adopted['reference'])).table()
+                        verboseprint("Old adopted measurement unset\n", old_adopted_data)
```
dr-rodriguez, line 780: Isn't this effectively the same as `old_adopted` above? Or do you have this here so you can check that adopted was reset?

scripts/ingests/utils.py
```diff
@@ -721,26 +741,57 @@ def ingest_parallaxes(db, sources, plxs, plx_errs, plx_refs, save_db=False, verb
...
-        parallax_data = [{'source': db_name,
+        if not duplicate:
+            # Construct data to be added
+            parallax_data = [{'source': db_name,
                           'parallax': str(plxs[i]),
                           'parallax_error': str(plx_errs[i]),
                           'reference': plx_refs[i],
                           'adopted': adopted}]
+
         verboseprint(parallax_data, verbose=verbose)
```
dr-rodriguez, line 795: You shouldn't need to convert `plx[i]` and similar to strings; leaving them as floats should be fine and would avoid issues were you may have values that can't become strings.

scripts/ingests/Gaia/ingest_gaia.py
```diff
@@ -0,0 +1,362 @@
...
+                      names=('db_names', 'designation', 'source_id'))
+
+    return result_table
+
+
+# Use SIMBAD to find Gaia DR2 and DR3 designations for sources without a Gaia designation
+# dr2_designations = find_in_simbad(no_gaia_source_id, 'Gaia DR2', source_id_index=2, verbose=VERBOSE)
+dr2_desig_file_string = 'scripts/ingests/Gaia/gaia_dr2_designations_'+DATE_SUFFIX+'.xml'
+# dr2_designations.write(dr2_desig_file_string, format='votable', overwrite=True)
+dr2_designations = Table.read(dr2_desig_file_string, format='votable')
```
dr-rodriguez, line 96: In development/testing, I understand why you would want to store this to a file, but I don't think this is something that needs to be added to the repo. The Gaia designations for DR2 are not going to change if you run this today, and if this is meant to be re-run we would want to re-create it anyways since we may have more sources in the SIMPLE database. In short: because these xml are dynamically drawn from external sources I would argue that we not store the xml results in this repo and instead have the code to regenerate them (which you do here).

scripts/ingests/Gaia/ingest_gaia.py
```diff
@@ -0,0 +1,362 @@
...
+        min_angdist_index = group['angular_distance'].tolist().index(min(group['angular_distance']))
+
+        if min_angdist_index == min_mag_index:
+            print('best one found \n', group['dr2_source_id', 'dr3_source_id'][min_angdist_index], '\n')
+            gaiadr3_unique.add_row(group[min_mag_index])
+        elif np.ma.is_masked(min(abs(group['magnitude_difference']))):
+            print('best one found just with angular distance \n', group['dr2_source_id', 'dr3_source_id'][min_angdist_index], '\n')
+            gaiadr3_unique.add_row(group[min_mag_index])
+        else:
+            print('no choice for \n', group['dr2_source_id', 'dr3_source_id', 'magnitude_difference', 'angular_distance'], '\n')
```
dr-rodriguez, line 168: You should add a few comment lines as to what this is doing, my understanding is that it is finding the best match in cases were there were multiple DR3 sources for the same DR2 source.
I assume magnitude_difference and angular_distance are things returned by the ADQL query above? If so, you might be able to modify the query to include something like `ORDER BY angular_distance DESC, magnitude_difference DESC` and that means the first row result would be likely to be the best match. Though how this works with your `group_py` approach, I'm not sure.

scripts/ingests/Gaia/ingest_gaia.py
```diff
@@ -0,0 +1,362 @@
...
+                     'width': 4053.},
+                    {'band': 'GAIA3.Grp',
+                     'instrument': 'Gaia',
+                     'telescope': 'Gaia',
+                     'effective_wavelength': 7620.,
+                     'width': 2924.}
+                    ]
+
+    db.PhotometryFilters.insert().execute(gaia_filters)
+    db.PhotometryFilters.insert().execute(gaia_dr3filters)
```
dr-rodriguez, line 251: An alternative to this would have been to modify https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/scripts/ingests/add_photometry_filters.py and add a few extra lines of the form:
```
{'telescope': 'Gaia', 'instrument': 'Gaia', 'filter': 'G'},
{'telescope': 'Gaia', 'instrument': 'Gaia', 'filter': 'Grp'},
```
since that scripts automatically adds missing Telescopes, Instruments, PhotometryFilters, if they don't exist, and it queries the SVO to get the wavelength/width information.
The only thing I see that would need verification is that it expects to construct the band name from the instrument+filter so the `fetch_svo` function would need some minor edits.

dr-rodriguez:
Overall, this looks fine. The comments I had were mainly minor suggestions or thoughts of alternative ways to do things. 

One final suggestion is that for your ingest script example, I would put all the functions up front near the top of the page so that the actual actions of ingest are not broken up by function definitions. That makes the workflow much neater.

dr-rodriguez (approved):
I'm approving this as-is, but encourage you to check the comments above for ways to improve things. I leave it up to you if you want to modify that for this PR or just consider it for future scripts.

scripts/ingests/utils.py
kelle: Yep, it is a check to see that the adopted flag has changed.

scripts/ingests/utils.py
kelle: If we don't convert to strings, it adds numbers to the values. e.g, 45.23 becomes something like 45.23000123. But since the datatype is float in the database, it is ingested as a float even when I convert to a string. So I prefer to do the string thing so that we don't end up with the extra decimals.

scripts/ingests/Gaia/ingest_gaia.py
kelle: Huh, ok, I will keep this is mind for future ingests.

scripts/ingests/Gaia/ingest_gaia.py
kelle: I added some comments but like the way that I select the best match explicitly.

scripts/ingests/Gaia/ingest_gaia.py
kelle: I removed the XML files from the PR.

scripts/ingests/utils.py
kelle: Ah, excellent -- thanks! Implemented.

scripts/ingests/utils.py
kelle: agreed.

PR #188, opened 2021-09-14
Utils Upgrade

Will-Cooper:
Adding SimpleError to be raised instead of print messages plus fixed some PEP compliance issues


Will-Cooper:
May need to rewrite the tests because naturally it's failing on the raised SimpleError

Will-Cooper:
Okie dokie, cc @kelle -- checks passed now

kelle (approved):

PR #194, opened 2021-09-22
Modified JSON files with Gaia data

kelle:
Adds 

- JSON files have Gaia proper motion, parallax, and photometry data.  
- Data tests of parallax, proper motion, and photometry measurements.

Also fixes some bugs in the `adopted` code in `ingest_parallaxes` and `ingest_proper_motion`


dr-rodriguez:
Looks good; loaded it locally and found no issues. I did notice a large number of significant figures (eg, `"parallax": 23.37133900094602`) which I thought were meant to be resolved by the use of `str()` in one of the ingest scripts?
Regardless, this is not a blocker so I'll approve this.

dr-rodriguez (approved):

kelle:
GAH!  I thought you wanted me to get rid of the string conversion!  I really don't like the extra digits so I'm re-doing this with the string conversion.

kelle:
turns out all those significant figures are actually in the Gaia data. So MERGING.

PR #195, opened 2021-09-23
Changing verboseprint into logger

Will-Cooper:
Now the only changes are so when you create a new function have a line saying:
```python
verbose = VerbosePrint(verbose)
```
then you can use
```python
verbose.print('your', 'message')
```


kelle:
Sorry, could you be more explicit?  Every function needs the 'verbose = VerbosePrint(verbose)' line in it, correct? 

And how would I use this for `table.pprint_all()`? What I was doing before, which stopped working, was:
```
if verobse:
     table.pprint_all()

**Comment by Will-Cooper** [comment]
> Sorry, could you be more explicit? Every function needs the 'verbose = VerbosePrint(verbose)' line in it, correct?

Yep precisely that, if every function is constructed with the keyword argument `verbose=True/False` then within that function you create an instance of the class `VerbosePrint` with `verbose = VerbosePrint(verbose)`, which constructs the class instance (`verbose`) using the boolean value of `verbose` from the function keywords.
Then whenever you want to print something just call `verbose.print('string')` as you would treat `print` normally -- however it will only actually print if the class instance was initialised with `verbose=True`.

> And how would I use this for `table.pprint_all()`? What I was doing before, which stopped working, was:
> 
> ```
> if verobse:
>      table.pprint_all()
> ```

I've added this in, similarly to `verbose.print` above, just call `verbose.pprint_all(<table_name>)` -- that'll only call the astropy table method `pprint_all` on the provided `<table_name>` if the instance was initialised with `verbose=True`.

**Comment by dr-rodriguez** [comment]
I don't like us using `verbose` to both be a variable of type boolean and a class/object name. It's far too easy to get mixed up and think verbose is True/False when it is sometimes an object. 
Can one use the old verboseprint and do something like `print = verboseprint(verbose=True)`? Wouldn't that be easier as it overwrites print to be the new verboseprint function?

**Comment by Will-Cooper** [comment]
Possibly, and I agree for sure with having a different named variable: one for the Boolean switch, one for the instance. 
I also like the idea of having `pprint_all` as a method so one doesn't have to repeatedly use
```python 
if verbose:
    table.pprint_all()
```

Will-Cooper:
huh the comment for that last commit is supposed to say:
rename `verbose` switch to `isverbose`

Will-Cooper:
Convert this into a [logger](https://docs.python.org/3/howto/logging.html#logging-basic-tutorial) function with debug and info levels

Will-Cooper:
I noticed in the tests that the logger messages weren't appearing, not sure what scripts to run to try the different levels e.g.
```python
logger.getLogger().setLevel = <logger.DEBUG/logger.INFO>
```
Essentially, having someone able to dynamically change the level of the logging whilst pulling in `utils` functions; ideas @dr-rodriguez?

Related: Any scripts which use the utils functions which change the kwarg `verbose` will need updating as I removed that from the functions.

scripts/ingests/utils.py
```diff
@@ -194,26 +184,27 @@ def add_names(db, sources=None, other_names=None, names_table=None, verbose=True
...
+            raise RuntimeError()
 
         for source, other_name in zip(sources, other_names):
             names_data.append({'source': source, 'other_name': other_name})
 
     if names_table is not None:
         if len(names_table[0]) != 2:
-            raise RuntimeError("Each row should have two elements")
+            logging.critical("Each row should have two elements")
+            raise RuntimeError()
```
dr-rodriguez, line 208: raise RuntimeError should still include the same text. So maybe create a variable for the text, `msg=Each row...` and then pass that to logging.error(msg) and raise RuntimeError(msg).
Also: don't use logging.critical, stick to debug/info/warning/error; critical should be very rare

Will-Cooper: Yeah I wasn't sure on the critical ones; I just assumed to put them in when something is about to cause the code to crash (e.g. an error about to be raised)?

scripts/ingests/utils.py
```diff
@@ -19,32 +20,19 @@
 import numpy as np
 
 warnings.filterwarnings("ignore", module='astroquery.simbad')
+logging.basicConfig(format='%(asctime)s %(levelname)s:%(message)s',
+                    datefmt='%m/%d/%Y %I:%M:%S %p')
```
dr-rodriguez, line 25: A good addition here is something like `logger = logging.getLogger('SIMPLE')`, you would then call logger.debug, etc instead of logging and if you need to refer to the logging level, you don't need the call to get the logger.

scripts/ingests/utils.py
dr-rodriguez: Personally, I actually use critical for things I always want to show up, regardless of whether it's an error or not, but that's probably not the best practice.

scripts/ingests/utils.py
```diff
@@ -103,15 +97,15 @@ def sort_sources(db, ingest_names, ingest_ras, ingest_decs, search_radius=60., v
     db_names = []
 
     for i, name in enumerate(ingest_names):
-        verboseprint("\n", i, ": searching:,", name, verbose=verbose)
+        logger.debug(''.join(map(str, ["\n", i, ": searching:,", name])))
```
dr-rodriguez, line 100: I recommend we switch these to f-strings, for example:
```
logger.debug(f"{i}: searching: {name}")
```
(I removed the \n since I don't think we really need that here)

This also applies to the other uses of join/map/str.
F-strings are a feature since python 3.6 and AstrodbKit2 has a minimum dependency of 3.7 or later so there is no reason for us not to use them.

dr-rodriguez (changes requested):
While this looks good, I recommend we avoid using `''.join(map(str, []))` We don't need the added complexity for what we're doing and it makes the code far less readable. Stick to using f-strings, which we already had in some cases.

scripts/ingests/utils.py
Will-Cooper: Oh yeah I totally agree and I love f strings normally; there were just 107 different print statements, most of which in the multiple args form; which was breaking logger (needed to be one single string); and I **really** didn't want to go through one by one to fix them... the `''.join(map(str, []))` method meant I could change every print statement at once thanks to pycharm...

PR #197, opened 2021-09-29
2MASS JSON files

kelle:
lots of modified JSON files with 2MASS designations and photometry. Also tests for new photometry measurements.


dr-rodriguez (approved):
Loaded these locally on my end without issues.

PR #213, opened 2021-11-10
Manja14 Spectral Ingest 

SherelynA:


kelle:
I figure out why the spectra aren't being ingested!  It's because the telescope, instrument, and modes need to be added. @SherelynA , this is your next step. Add ISSAC, SW LRES, and ESO VLT U3.

[ISAAC was on VLT U3](https://www.eso.org/sci/facilities/paranal/decommissioned/isaac.html) so we should add "ESO VLT U3" as a telescope rather than just "VLT". (VLT is actually 4 different telescopes.)

data/2mass_j12451416-4429077.json
```diff
@@ -243,6 +246,20 @@
             "wavelength_order": null,
             "comments": null,
             "reference": "Loop07b"
+        },
+        {
+            "spectrum": "\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000",
```
kelle, line 251: Uh-oh. This is a problem. Not sure what's going on here. What's the spectrum supposed to be?

data/cha_130540-773958.json
```diff
@@ -14,6 +14,25 @@
...
+            "local_spectrum": null,
+            "regime": "nir",
+            "telescope": "ESO VLT U3",
+            "instrument": "ISAAC",
+            "mode": "SW LRes",
+            "observation_date": "2010-04-21T00:00:00",
+            "wavelength_units": null,
+            "flux_units": null,
+            "wavelength_order": null,
+            "comments": "April 6, 2010",
```
kelle, line 34: The comment with an additional observation date should have more words. For example, "spectrum also includes data observed on April 6, 2010."

kelle (changes requested):

data/cha_130540-773958.json
SherelynA: I made the comment more comprehensive in the new csv.

kelle (approved):

PR #216, opened 2021-11-12
Ingest one Y dwarf. Closes #190.

### LINKED ISSUE #190: Ingest one Y dwarf
Ingest the one Y dwarf which didn’t get ingested by Clem’s earlier script because it’s not in Simbad.

kelle:
Ingests the one Y dwarf from Kirk19 which wasn't/isn't in SIMBAD.  Also includes updates to `ingest_utils` to better handle ingesting one source.

### DIFF
--- a/data/wisea_j030237.52-581740.3.json
+++ b/data/wisea_j030237.52-581740.3.json
@@ -16,6 +16,39 @@
             "other_name": "WISEA J030237.52-581740.3"
         }
     ],
+    "Parallaxes": [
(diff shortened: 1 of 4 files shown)


kelle:
Looking at this more closely, this didn't actually ingest the source. The source must have been ingested between then and now. But this does add the spectral type, Spitzer photometry, and astrometry from Kirk19.

dr-rodriguez (approved):

PR #220, opened 2021-11-17
Fix to some publications

kelle:
Fixes Martin duplicate. Will also work on Schm10b duplicate.

### DIFF
--- a/simple/schema.py
+++ b/simple/schema.py
@@ -13,7 +13,7 @@ class Publications(Base):
     This stores reference information (DOI, bibcodes, etc) and has shortname as the primary key
     """
     __tablename__ = 'Publications'
-    name = Column(String(30), primary_key=True, nullable=False)
+    name = Column(String(30), primary_key=True, nullable=False, onupdate='cascade')
     bibcode = Column(String(100))
     doi = Column(String(100))
     description = Column(String(1000))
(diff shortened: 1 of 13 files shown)


kelle:
Two things:

**1) Cascade update good**

   When I update the reference name in the Publications table, it cascades up to the Sources table IF the Sources table is the only table that uses that particular reference. E.g., when I change the name in the Publications table, it automagically changes the reference in the Sources table.

   HOWEVER, if that reference is also used in the Proper Motions table, I get an integrity error. I don't understand why the change in the Publications table doesn't also "cascade" to the Proper Motions table.

   Bottom line is that I don't understand `ondelete` and `onupdate`. 

**2) Cascade delete bad**

   ALSO, when I deleted a record from Publications when it was being used in the Proper Motion's table, it deleted the proper motion measurements. I think this is *not* the behavior we want but I couldn't figure out how to prevent it from happening. 

It seems like I both need to learn more and that we may need to tweak the schema a bit more.

simple/schema.py
```diff
@@ -13,7 +13,7 @@ class Publications(Base):
     This stores reference information (DOI, bibcodes, etc) and has shortname as the primary key
     """
     __tablename__ = 'Publications'
-    name = Column(String(30), primary_key=True, nullable=False)
+    name = Column(String(30), primary_key=True, nullable=False, onupdate='cascade')
```
dr-rodriguez, line 16: You don't need onupdate here, since that is only for ForeignKey relations.

dr-rodriguez (approved):

PR #221, opened 2021-11-17
JSON updates of Schm10 references. Closes #168

### LINKED ISSUE #168: Fix Schm10b Publication Record
Publication Schm10b is in the database twice, both as Schm10 and Schm10b.

kelle:


Will-Cooper:
Have we documented this somewhere? Changing duplicate short refs to be NameYY.DOI? Or however it is we're doing this?

simple/schema.py
```diff
@@ -13,7 +13,7 @@ class Publications(Base):
     This stores reference information (DOI, bibcodes, etc) and has shortname as the primary key
     """
     __tablename__ = 'Publications'
-    name = Column(String(30), primary_key=True, nullable=False)
+    name = Column(String(30), primary_key=True, nullable=False, onupdate='cascade')
```
Will-Cooper, line 16: what is this doing?

Will-Cooper (commented):
Python looks good

dr-rodriguez (approved):

simple/schema.py
dr-rodriguez: I will un-do this after the PR since that shouldn't be there

PR #222, opened 2021-11-19
Add codepath to use fast load_db when supported by astrodbkit2

blackwer:
This fix will attempt to use the new in-memory database functionality in `astrodbkit2==0.2.6` in the `load_db` function to create databases, and will fall back to the old file based method if that fails.


kelle (approved):
Good catch!

PR #223, opened 2021-11-30
Updating schema to use cascade updates

dr-rodriguez:
Updating schema to have correct use of cascade publication changes on updates. Some tables had incorrectly set to on delete which causes surprising issues (a lot of data gets deleted if a publication is deleted- we only want that for sources)


kelle (approved):
LGTM!

PR #226, opened 2021-12-08
Ingesting spectra for Manj16

SherelynA:


kelle:
Xshooter has only one mode so providing a mode isn't necessary.

SherelynA:
Everything seems to looks good and ready to merge

kelle:
This looks great!  Only thing which we should do is to add a new data test for the `em.IR.NIR` and `em.opt` regimes. 

On lines 243-249 of `test_data.py` there are tests for `nir` and `optical`. Can you use those a guide to add two new tests for  `em.IR.NIR` and `em.opt` regimes?

kelle (changes requested):
This looks great!  Only thing which we should do is to add a new data test for the `em.IR.NIR` and `em.opt` regimes. 

On lines 243-249 of `test_data.py` there are tests for `nir` and `optical`. Can you use those a guide to add two new tests for  `em.IR.NIR` and `em.opt` regimes?

SherelynA:
> This looks great! Only thing which we should do is to add a new data test for the `em.IR.NIR` and `em.opt` regimes.
> 
> On lines 243-249 of `test_data.py` there are tests for `nir` and `optical`. Can you use those a guide to add two new tests for `em.IR.NIR` and `em.opt` regimes?

I added the two new test that check for those two regimes.

kelle (approved):

kelle:
This looks perfect -- great work!

PR #232, opened 2021-12-20
Manj19 JSON files

kelle:
Added `other_references` column to Sources and Spectra tables so all JSON files were modified. Also added 77 spectra from Manj19.


dr-rodriguez:
May want to explore the Provenance data model at https://www.ivoa.net/documents/ProvenanceDM/20200411/index.html to see if we can implement a provenance table to store the history of how a file got processed.

Will-Cooper (commented):
Do we have checks that the spectra load/ work?

PR #240, opened 2022-01-25
JSON files for Manj19 ingest

kelle:


kelle:
The wavelength and flux units aren't in the JSON files but should be.

dr-rodriguez (approved):

PR #246, opened 2022-04-08
Tests/ Requirements for Publications Table

### LINKED ISSUE #98: Write tests/requirements for the Publications table
shortname is required, but should bibcode and/or doi be required? What about sources which are actually unreferenced or unpublished?

SherelynA:
- [x]  Find unused references
- [x]  Find references with no DOI and no Bibcode and no description
        closes #98

### DIFF
--- a/tests/test_integrity.py
+++ b/tests/test_integrity.py
@@ -1,9 +1,11 @@
 # Test to verify database integrity
 
 import os
+from operator import and_
+
 import pytest
 from . import REFERENCE_TABLES
-from sqlalchemy import func
+from sqlalchemy import func, select, except_
 from simple.schema import *
 from astrodbkit2.astrodb import create_database, Database, or_
 from astropy.table import unique
@@ -82,8 +84,22 @@ def test_references(db):
         print(f'\n{len(t)} publications not referenced by {table_list}')
         print(t)
 
+
 def test_publications(db):
- #     find unused references
+    # Find unused references in the Sources Table
+    stm = except_(select([db.Publications.c.publication]), select([db.Sources.c.reference]))
+    result = db.session.execute(stm)
+    s = result.scalars().all()
+    assert len(s) == 720, f'found {len(s)} unused references'
+
+    # Find references with no doi or bibcode
+    t = db.query(db.Publications.c.publication).filter(
+        or_(and_(db.Publications.c.doi.is_(None), db.Publications.c.bibcode.is_(None)),
+            and_(db.Publications.c.doi.is_(''), db.Publications.c.bibcode.is_(None)),
+            and_(db.Publications.c.doi.is_(None), db.Publications.c.bibcode.is_('')),
+            and_(db.Publications.c.doi.is_(''), db.Publications.c.bibcode.is_('')))).astropy()
+    assert len(t) == 26, f'found {len(t)} publications with missing bibcode and doi'
+
 
 def test_coordinates(db):
     # Verify that all sources have valid coordinates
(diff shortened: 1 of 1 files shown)


tests/test_integrity.py
```diff
@@ -83,6 +85,22 @@ def test_references(db):
...
+    s = result.scalars().all()
+    assert len(s) == 720, f'found {len(s)} unused references'
+
+    # Find references with no doi or bibcode
+    t = db.query(db.Publications.c.publication).filter(
+        or_(and_(db.Publications.c.doi.is_(None), db.Publications.c.bibcode.is_(None)),
+            and_(db.Publications.c.doi.is_(''), db.Publications.c.bibcode.is_(None)),
+            and_(db.Publications.c.doi.is_(None), db.Publications.c.bibcode.is_('')),
+            and_(db.Publications.c.doi.is_(''), db.Publications.c.bibcode.is_('')))).astropy()
+    assert len(t) == 26, f'found {len(t)} publications with missing bibcode and doi'
```
kelle, line 101: This looks great!

tests/test_integrity.py
```diff
@@ -83,6 +85,22 @@ def test_references(db):
         print(t)
 
 
+def test_publications(db):
+    # Find unused references in the Sources Table
+    stm = except_(select([db.Publications.c.publication]), select([db.Sources.c.reference]))
+    result = db.session.execute(stm)
+    s = result.scalars().all()
+    assert len(s) == 720, f'found {len(s)} unused references'
```
kelle, line 93: Keep this code -- we might want it for something else in the future!  But it's not doing exactly what we want it to. Instead, work on modifying the above test (`test_references`) to include more tables.

kelle, line 93: Keep this code -- we might want it for something else in the future!  But it's not doing exactly what we want it to. Instead, work on modifying the above test (`test_references`) to include more tables.

kelle (changes requested):

tests/test_integrity.py
```diff
@@ -75,14 +77,32 @@ def test_references(db):
     # Confirm that all are in Publications
     t = db.query(db.Publications.c.publication).filter(db.Publications.c.publication.in_(ref_list)).astropy()
     assert len(t) == len(ref_list), 'Some references were not matched'
+    assert len(t) == 578, f'{len(t)} publications referenced'
```
kelle, line 80: Let's think about this. Do we really want to modify this test every time we add a reference?

tests/test_integrity.py
```diff
@@ -75,14 +77,32 @@ def test_references(db):
...
 
     # List out publications that have not been used
     t = db.query(db.Publications.c.publication).filter(db.Publications.c.publication.notin_(ref_list)).astropy()
+    assert len(t) == 568, f'{len(t)} unused references'
```
kelle, line 84: That's a lot. Hm.

tests/test_integrity.py
```diff
@@ -75,14 +77,32 @@ def test_references(db):
...
     if len(t) > 0:
         print(f'\n{len(t)} publications not referenced by {table_list}')
         print(t)
```
kelle, line 87: we should comment this out.

tests/test_integrity.py
kelle: Lets change to less than or equal to 568.

tests/test_integrity.py
kelle: No, lets get rid of it.

kelle (changes requested):

kelle (approved):
Nice work!

PR #252, opened 2022-06-08
Spectral Type Ingest for 2020 Best Sample

SherelynA:


scripts/ingests/2020Best_spectraltype_ingest.py
```diff
@@ -0,0 +1,110 @@
...
+opt_names = filter_opt_data['name']
+n_sources_opt = len(opt_names)
+opt_regime = ['optical'] * n_sources_opt
+spt_refs_opt = filter_opt_data['ref_spt_opt']
+spectral_types_opt = filter_opt_data['spt_opt']
+# IR Spectral Types
+filter_ir_data = data[data['spt_ir'] != 'null']
+ir_names = filter_ir_data['name']
+n_sources_ir = len(ir_names)
+ir_regime = ['nir_UCD'] * n_sources_ir
```
kelle, line 31: Hm, something is wrong here. it's coming up as `nir_UCD` and I think we just want `nir` but I don't remember exactly.

dr-rodriguez:
This has been updated to fix the reference matching and includes new tests. Best to do after closing PR #270 as it includes the changes from there.

scripts/ingests/2020Best_spectraltype_ingest.py
dr-rodriguez: Related to #271 but something we can do cleanup in a separate PR

kelle (approved):

PR #267, opened 2022-07-22
Fixing duplicate publications

dr-rodriguez:
While looking into #252 I discovered several cases where multiple publications share the same bibcode. This is incorrect so I effectively merged the two entries together.
I also took the liberty of removing accents from Gome01 and Lope04 to avoid the unicode issues we have been seeing.


dr-rodriguez:
Because this merges fix_unicode, this includes work done as part of PR #268

data/2mass_j00325584-4405058.json
```diff
@@ -7,7 +7,7 @@
             "epoch": null,
             "equinox": null,
             "shortname": "0032-4405",
-            "reference": "Gold99",
+            "reference": "Eros99c",
```
kelle, line 10: we shouldn't be using "c".  Change to Eros99.

scripts/updates/fix_duplicate_publications.py
```diff
@@ -0,0 +1,55 @@
...
+
+for bibcode in t['bibcode']:
+    find_publication(db, bibcode=bibcode)
+
+# ---------------------------------------------------------------------------
+# Manual fixes to publications
+
+# This transformation dictionary contains the references that are to be converted
+# Additional entries can be added
+to_transform = {'Gold99': 'Eros99c',
```
kelle, line 26:
```suggestion
to_transform = {'Gold99': 'Eros99',
```

scripts/updates/update_publications.py
```diff
@@ -17,6 +17,10 @@
 db.Parallaxes.update().where(db.Parallaxes.c.reference == 'Liu16').values(reference='Liu_16').execute()
 db.Publications.delete().where(db.Publications.c.publication == 'Liu16').execute()
 
+# Removing accents
+db.Publications.update().where(db.Publications.c.publication == 'Góme01').values(publication='Gome01').execute()
+db.Publications.update().where(db.Publications.c.publication == 'Lópe04').values(publication='Lope04').execute()
```
kelle, line 22: I'm not convinced. I kinda want to leave these. The accents aren't causing any problems.

scripts/updates/update_unicode.py
```diff
@@ -0,0 +1,33 @@
+# Script to address unicode issues from some incorrectly inserted data
+import logging
+from scripts.ingests.utils import logger, load_simpledb
+
+logger.setLevel(logging.INFO)
+
+db = load_simpledb('SIMPLE.db', recreatedb=True)
+
+# Character to fix:
+data = db.search_string('\u2013')
```
kelle, line 10:
```suggestion
# Character to fix:
data = db.search_string('\u2013') # \u2013 Corresponds to an en-dash. Change to regular dash.
```

kelle: I also propose a test which is that the results of this search should be empty.

kelle (changes requested):
awesome progress. Thanks for the scripts!

scripts/updates/update_publications.py
dr-rodriguez: The accents are causing problems- they can flip to unicode designations in some machines depending on the order of operations (which is what happened on the other PR)

kelle: ok, I won't die on this hill.

tests/test_integrity.py
```diff
@@ -490,6 +490,15 @@ def test_spectra(db):
     assert len(flux_unit_fail) == 0, f'Some flux units did not resolve: {flux_unit_fail}'
 
 
+def test_special_characters(db):
+    # This test asserts that no special unicode characters are in the database
+    # This can be expanded with additional characters we want to avoid
+    bad_characters = ['\u2013']
```
kelle, line 496: Lets also add ó since we you just got rid of it! And another common on é
\u00f3 ó
\u00e9 é
```suggestion
    bad_characters = ['\u2013', '\u00f3', '\u00e9']
```

dr-rodriguez: It looks a little more complex now, since I wanted to avoid failing on descriptions and some paper titles do have accents, but this is working now. Includes a,e,i,o,u

kelle (approved):
Looks great!!!

PR #283, opened 2022-08-23
Versions update for generation script

dr-rodriguez:
The automatic database generation script overwrites the REFERENCE_TABLES and that causes issues (which means SIMPLE-web will not have this version). This simple update makes this more flexible.


dr-rodriguez:
Something is broken in the database generation with the Versions table; debugging...

dr-rodriguez:
Ended up removing the kwargs statements from load_simpledb. That in addition to the reference table use was messing up foreign key checks everywhere.

kelle (approved):

Issue #271, opened 2022-07-28
Clean up regime values in database and make reference table


dr-rodriguez:
We have a mismatch of regime values across the database and in the enumeration description, we have things like optical and optical_UCD. I don't think UCD are really adding anything to this and are just causing inconsistencies so I propose we remove all UCD regime values and turn them to their simpler counterparts; so instead of using things like `em.IR.NIR` we just use `nir`

Here are the values (post the Best2020 SpectralType PR) we have:
```
t = db.query(db.Spectra.c.regime, func.count(db.Spectra.c.regime).label('counts')). \
    group_by(db.Spectra.c.regime). \
    astropy()
print('Spectra')
print(t)

t = db.query(db.SpectralTypes.c.regime, func.count(db.SpectralTypes.c.regime).label('counts')). \
    group_by(db.SpectralTypes.c.regime). \
    astropy()
print('SpectralTypes')
print(t)

Spectra
  regime  counts
--------- ------
em.IR.NIR    118
   em.opt     21
      mir     91
      nir    456
  optical    718
SpectralTypes
 regime  counts
-------- ------
infrared     22
     nir     41
 nir_UCD   2081
 optical   1494
 unknown     10
```

kelle:
This is a follow-up to #205. And I'll put a proposed revised table:
regime | UCD  | Description
-- | -- | --
opt | em.opt | Optical part of the spectrum
J | em.IR.J | Infrared between 1.0 and 1.5 micron
H | em.IR.H | Infrared between 1.5 and 2 micron
K | em.IR.K | Infrared between 2 and 3 micron
NIR | em.IR.NIR | Near-Infrared, 1-5 microns
MIR | em.IR.MIR | Medium-Infrared, 5-30 microns
FIR | em.IR.FIR | Far-Infrared, 30-100 microns

dr-rodriguez:
Thanks, I didn't realize we had a similar issue already in place.
I think there are 2 things to consider here: one is the consolidation of our different values, but the other is the use of UCD here.

[...]

PR #292, opened 2022-10-04
Implementation of ModeledParameters table

### LINKED ISSUE #244: Make Modeled/Derived Parameter Table
Make a long table for derived and modeled parameters. first ingest can be the distances from Bailer Jones 2021 (#243 ).
Columns:
- Source
- Parameter - constrained name space.
- Value
- Low value
- High value
- maximum likelihood 
- mean marginalized posterior
- variance of marginalized posterior
- Comment - could include what "value", "low", and "high" correspond to.
- Reference

dr-rodriguez:
Initial implementation of ModeledParameters table.
This is fairly simple and includes the enumeration constraints for that table. I have encountered some issues with enumerations so I'm trying to keep them as simple as possible (see #271)

Do we want to create a test that loads up some fake ModeledParameters data?

Closes #244


simple/schema.py
```diff
@@ -97,6 +98,16 @@ class Gravity(enum.Enum):
...
 
 
+class Parameter(enum.Enum):
+    """Enumeration for derived/inferred parameters"""
+    mass = 'mass'
+    radius = 'radius'
+    log_g = 'log_g'
+    T_eff = 'T_eff'
+    metallicity = 'metallicity'  # Z
+    CO = 'CO'  # C/O
```
kelle, line 108: maybe we should be more pedantic here. I think `CO` is too ambiguous/misleading. Maybe `C/O ratio`? I dont think there's a problem with the slash in the string is there? If so, `C-to-O ratio`?

dr-rodriguez: I need to confirm (and maybe write a test for) but I'm 80% sure that what gets stored in the database is the variable, not the string. Which means things like `C/O` and `C-to-O` are invalid.

kelle: C_to_O_ratio

kelle:
Let's not use an enumeration for Parameter's, let's use a reference table.

simple/schema.py
dr-rodriguez: Switching to an actual table, so the enumeration question is no longer relevant

dr-rodriguez:
Explicitly populate the Parameters table with: mass, radius, log_g, T_eff, metallicity, C-to-O ratio.

dr-rodriguez:
Should Parameters have units? If so, what should we store for these? (I imagine that some are NULL, which is fine)

kelle (changes requested):

kelle (approved):

PR #300, opened 2022-10-14
Fix to PhotometryFilters

kelle:


kelle:
This adds all used Photometry bands to the PhotometryFilters and adds a relevant test.

There are more improvements which could be done (e.g., add References to the new Instruments, add more UCDs to the Photometry table, make sure all filters are in the SVO format) but this at least fixes the problem at hand.

Will-Cooper (approved):
Looks good

PR #302, opened 2022-10-17
Update gen-db.yml

Will-Cooper:
only trigger db build on *published* release, see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release


kelle (approved):
Nice!

PR #305, opened 2022-10-18
Plot examples

dr-rodriguez:
Some basic examples of queries against SIMPLE and plotting the results with matplotlib. Also includes a function to reverse the spectral_type_code into a string, for ease of grouping results.


kelle (approved):
Gorgeous plots! 

The one thing we're not user about is why you converted the code to a string. You could have just used the `spectral_type_string` field. But I suspect you had a reason!

Will-Cooper (approved):
LGTM

dr-rodriguez:
spectral_type_string contains more information than just L3 or M5, there are peculiarities stored there which makes grouping by that more difficult. Maybe an argument too much information is being stored in that column.

PR #318, opened 2022-11-12
Set up for GitHub Codespaces

dr-rodriguez:
This is a quick setup for preparing a GitHub Codespace environment with python 3.8 and a local postgres server. I can walk us through how to use it, but it allows us to spin up a standard development environment with everything configured for using SIMPLE.

The postgres setup is a bonus- the DB still has some issues so we need a separate PR to address them, but it serves as proof-of-concept.


kelle (approved):
fancy!

PR #331, opened 2022-12-28
Create pull request template

kelle:
Based on @dr-rodriguez astute observations, there's a lot to remember to do when doing a data ingest pull request. Maybe we want to use a template to help us remember.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,6 @@
-Fixes #
+Short description: Include what type of data being ingested and appropriate references.
+
+Link to relevant issue: Closes #
 
 For data ingests:
 [ ] includes script used for ingest
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+Fixes #
+
+For data ingests:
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,7 @@
+Fixes #
+
```
dr-rodriguez, line 2: There should be a `Short Description:` section to highlight what the main goal is of the PR, beyond what's in the title

.github/pull_request_template.md
```diff
@@ -0,0 +1,7 @@
...
+For data ingests:
+[ ] includes script used for ingest
+[ ] includes modified JSON files
```
dr-rodriguez, line 5: At one point we were doing JSON files as part of a second PR, do we want to still do that or keep it in the same PR? The advantage of it being together is that it's easier for the folks preparing the PR, the disadvantage is that there maybe a lot of files to review.

kelle: The 2nd PR for the JSON files seems silly. I've just been getting the ingest script reviewed before generating the JSON files.

dr-rodriguez (approved):

Issue #325, opened 2022-11-22
Units are resolvable by astropy


dr-rodriguez:
We want units in the Modeled Parameters table to be resolvable by the [astropy.units module](https://docs.astropy.org/en/stable/api/astropy.units.Unit.html#astropy.units.Unit).

Ajb2307:
Hi, I want to start working on this. I see it's labeled as tests. Is that where you want to script for this?

kelle:
The test should go in the `tests/test_integrity.py` module.  It should actually be very similar to the test already in `test_spectra` function starting on line 469 of that file. You can use that as a guide to help get started. 

One challenge is that there is no data in the modeled parameters table yet so it will be hard to see if the test is working but I'm all in favor of having the test in place before we start adding data!

Ajb2307:
Im going to ingest test parameters from Table 9 of Filippazzo 
https://ui.adsabs.harvard.edu/abs/2015ApJ...810..158F/abstract
To test code I've written

PR #340, opened 2023-02-06
Making astropy units resolvable on modeled parameters

Ajb2307:
Closes Issue #325  

- [x]  Creating code to test units that will go into modeled parameters

- [x]  Will test with an ingest from https://iopscience.iop.org/article/10.1088/0004-637X/810/2/158#apj518886t9 table 9

- [x]  Test incorrectly formatted data mark as expected fails


tests/test_integrity.py
```diff
@@ -476,6 +476,239 @@ def test_sources(db):
...
+        {'source': '2228-4310', 'value': '0.94 +or- 0.16', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ,
+        {'source': '2234+4041', 'value': 'None', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ,
+        {'source': '2237+3922', 'value': '1.04 +or- 0.06', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ,
+        {'source': '2244+2043', 'value': '1.29 +or- 0.03', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ,
+        {'source': '2306-0502', 'value': '1.14 +or- 0.04', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ,
+        {'source': '2322-3133', 'value': '1.39 +or- 0.19', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ,
+        {'source': '2322-6151', 'value': '1.59 +or- 0.05', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ,
+        {'source': '2351-2537', 'value': '1.16 +or- 0.11', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ,
+        {'source': '2354-3316', 'value': '1.09 +or- 0.05', 'parameter': 'radius', 'publication': 'Fili15', 'unit': '(R_Jup)'} ])  # change to csv
```
kelle, line 680: This is the beginning of the ingest. Just keep data for one source and put the rest into a CSV file and save for later.

Let's also add the other parameters for the same source.

tests/test_integrity.py
```diff
@@ -12,6 +12,7 @@
 from astropy import units as u
 from astroquery.simbad import Simbad
 from astrodbkit2.utils import _name_formatter
+from astropy.constants import R_jup
```
kelle, line 15: @dr-rodriguez , what do you think of this?  It makes the test pass but I'm not sure it's the best thing for us to do.

dr-rodriguez: Where is this used? From the changed lines the only place I see R_jup is as a string, not as a variable.

kelle: It's as a string in the ModeledParameters table test data but is then tested as a unit. It passes 
```python
try:
     assert u.Unit(unit, parse_strict='raise')
```

But then, in the tests, we have to explicitly import which constants we're using as "units".  I think this is ok as long as we indicate in the docs.

dr-rodriguez: Interesting behavior on astropy.unit's part. I guess that's OK, but we may want to import a few more constants (M_Jup at least).

kelle:
Great progress -- Awesome that all of those units are resolvable!  

Next step would be to add some data which have _incorrectly_ formatted units. Maybe try "Kelvins" and "M_jupiter".  Make some tests which fail and then on Friday we'll discuss how to mark them as "expected" to fail.

tests/test_integrity.py
Ajb2307: It turns out that M_jup and R_jup are resolvable through from astropy import units as u and no addition imports are needed

tests/test_integrity.py
```diff
@@ -476,6 +477,51 @@ def test_sources(db):
     assert len(t) == 37, f'found {len(t)} sources from {ref}'
 
 
+# importing table 9 for data in test parameters
+from astropy.table import Table
```
dr-rodriguez, line 481: Any import statements should be at the top of the file, for consistency and clarity.

tests/test_integrity.py
```diff
@@ -476,6 +477,51 @@ def test_sources(db):
...
+
+def test_ingest_modeledparameters(db):
```
dr-rodriguez, line 483: I think you got lucky that this ran before `test_modeledparameters`. I think maybe because it's alphabetically before? I don't know if pytest runs things sequentially or alphabetically, but if you want to guarantee that these inserts are in place before you run the tests either move them to `test_modeledparameters`, the `db` fixture at the start of the file, or as an extra fixture (probably not needed). 
Also, if you are inserting data, you could run into unrelated integrity errors if these sources are already present or the references aren't in place. Just a heads up to be clear on what you are testing.

Ajb2307: Thanks. I just deleted 481 as it's not actually needed. I had put it in a while ago and forgot

tests/test_integrity.py
```diff
@@ -476,6 +477,54 @@ def test_sources(db):
...
+        unit = x['unit']
+
+        try:
+            assert u.Unit(unit, parse_strict='raise')
+        except ValueError:
+            print(f'{unit} is not a recognized astropy unit')
+            counts = db.query(db.ModeledParameters).filter(db.ModeledParameters.c.unit == unit).count()
+            unit_fail.append({unit: counts})  # count of how many of that unit there is
+
+    assert len(unit_fail) == 0, f'Some parameter units did not resolve: {unit_fail}'
```
kelle, line 525: I think this is the line that may need to change give the units we expect to fail.  Maybe `len(unit_fail) == 4`.

tests/test_integrity.py
```diff
@@ -476,6 +477,54 @@ def test_sources(db):
...
+                         {'source': '2MASS J00001354+2554180', 'value': 5.02, 'parameter': 'log g', 'reference': 'Fili15', 'unit': 'dex'},
+                         {'source': '2MASS J00001354+2554180', 'value': 1227.0, 'parameter': 'T eff', 'reference': 'Fili15', 'unit': 'K'},
+                         {'source': '2MASS J00001354+2554180', 'value': 47.56, 'parameter': 'mass', 'reference': 'Fili15', 'unit': 'M_jup'},
+                         # next data points should fail
+                         {'source': '2MASS J00034227-2822410', 'value': 106.97, 'parameter': 'mass', 'reference': 'Fili15', 'unit': 'M_Jupiter'},
+                         {'source': '2MASS J00034227-2822410', 'value': 1.32, 'parameter': 'radius', 'reference': 'Fili15', 'unit': 'R_Jup'},
+                         {'source': '2MASS J00034227-2822410', 'value': 2871.0, 'parameter': 'T eff', 'reference': 'Fili15', 'unit': 'kelvin'},
+                         {'source': '2MASS J00034227-2822410', 'value': 5.18, 'parameter': 'log g', 'reference': 'Fili15', 'unit': '.'}]
+
+    db.ModeledParameters.insert().execute(model_params_data)
```
kelle, line 493: Move all of this INTO `test_modeled_parameters` function.

tests/test_integrity.py
```diff
@@ -476,6 +477,54 @@ def test_sources(db):
...
+
+
+
+
+def test_modeledparameters(db):
```
kelle, line 498:
```suggestion
def test_modeledparameters(db):
```
```suggestion
def test_modeled_parameters(db):
```

kelle (approved):
Looks great!!

PR #344, opened 2023-02-22
Update the author codes in the publications table

### LINKED ISSUE #171: Author Codes
For the author codes published in the same year, we should have some consistent way to a) make sure data contributors align the same paper (e.g. not just "I reference it as Burg05b, so I'll link it to the Burg05b in the SIMPLE database" even if that's not correct and b) something easy to check and look at.

I propose appending with perhaps the last two or three digits of the DOI, so very quick to cross check.

SherelynA:
This script will change the author codes in publication so that they are formatted in a more logical way in the table. Will use the format @kelle proposed in the #171 issue conversation. 
Closes #171


kelle:
Great idea - happy to help! Did you want to script this or do it by hand? I think it will be faster to do by hand...especially since it's a one-off task.


We should also update the documentation as part of this PR.

SherelynA:
@kelle I think it will be nice to do a a sql query to determine the publications that need to be changed so we can just change all at once those that need to be changed in one go. Also, yes we should update the documentation as part of the PR.

kelle (commented):
This looks good to me but I would want @dr-rodriguez to approve it also.

scripts/ingests/update_authorcodes.py
```diff
@@ -0,0 +1,54 @@
...
+            db.Publications.update().where(db.Publications.c.publication == publication).values(
+                publication=new_publication_name).execute()
+        elif bibcode[17:-1].find('.') == -1:
+            new_publication_name = publication[0:6] + '.' + bibcode[17:-1]
+            db.Publications.update().where(db.Publications.c.publication == publication).values(
+                publication=new_publication_name).execute()
+        elif bibcode[18:-1].find('.') == -1:
+            new_publication_name = publication[0:6] + '.' + bibcode[18:-1]
+            db.Publications.update().where(db.Publications.c.publication == publication).values(
+                publication=new_publication_name).execute()
```
dr-rodriguez, line 50: This is a lot of duplicate code just to construct a new publication name. You can do the same in much fewer lines:
```
ind = bibcode.rfind('.')  # find index of the last period
if ind != -1:   
    # if a match was found, generate the new name
    new_publication_name =  publication[0:6] + '.' + bibcode[ind+1:-1]
    db.Publications....  # run the db command
```

dr-rodriguez: Note the use of `rfind` as opposed to `find`: https://docs.python.org/3/library/stdtypes.html?highlight=rfind#str.rfind

scripts/ingests/update_authorcodes.py
```diff
@@ -0,0 +1,40 @@
...
+
+#  creating a query that finds all the author codes that need to be updated
+stm = select([db.Publications.c.publication, db.Publications.c.bibcode]).where(
+    or_(db.Publications.c.publication.like('______a'), db.Publications.c.publication.like('______b'),
+        db.Publications.c.publication.like('______c'), db.Publications.c.publication.like('______d'),
+        db.Publications.c.publication.like('______e'), db.Publications.c.publication.like('______f'),
+        db.Publications.c.publication.like('______g'), db.Publications.c.publication.like('______h')))
+
+result = db.session.execute(stm)
+df = pd.DataFrame(result.all())
```
dr-rodriguez, line 26: BTW, I realized that this can also be compacted a little bit, particularly if you use the ORM way to call it (as opposed to select). For example, something like:
```
df = db.query(db.Publications.c.publication, db.Publications.c.bibcode). \
     filter(or_(db.Publications.c.publication.like('______a'), 
                db.Publications.c.publication.like('______b'), 
                etc)). \
     pandas()
```
But I'll approve regardless- no need to change unless you really want to.

dr-rodriguez (approved):

PR #345, opened 2023-02-24
SQLAlchemy 2.0 support

### LINKED ISSUE #341: Handle SQLAlchemy 2.0 failures
Currently all tests are failing due to an update to SQL Alchemy to v2. @dr-rodriguez is going to investigate how to make `astrodbkit` work with this new version. Otherwise we may need to pin to a v1 of `sqlalchemy`.

dr-rodriguez:
Short description: Updating tests and ingest scripts for SQLAlchemy 2.0

Link to relevant issue: Closes #341

### DIFF
--- a/.github/workflows/gen-db.yml
+++ b/.github/workflows/gen-db.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install git+https://github.com/dr-rodriguez/AstrodbKit2.git@sqlalchemy2
+        pip install astrodbkit2
 
     - name: Generate sqlite (file) database
       run: |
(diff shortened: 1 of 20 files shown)


dr-rodriguez:
There were quite a few commits, but I squashed them to clean up the history a little. As it stands, this may require us to coordinate the Astrodbkit2 release. Version 0.4.2 (the latest) is pinned to avoid sqlalchemy 2.0; Version 0.5 (the upcoming) will use the updated sqlalchemy which these changes to the SIMPLE code base rely on.

kelle (approved):

kelle:
So I understand: Right now, we're using the latest Astrodbkit2 from Github but after version 0.5 is released, we can go back to installing from pip?

dr-rodriguez:
That's correct- once we release 0.5 we can change the github actions to use pip

PR #348, opened 2023-03-14
Remove spectra units.

### LINKED ISSUE #324: Remove Units Column from Spectra table
The original files for Suarez22 are in Janksy's and the converted files are in ergs/... Should we just get rid of the units column so this doesn't happen again, where we have mismatching units?

kelle:
Modifies schema and data files to remove flux, wavelength units and wavelength_order from the Spectra table.

Closes #324


dr-rodriguez (dismissed):

dr-rodriguez:
May need to update the documentation as well...

dr-rodriguez (approved):

PR #351, opened 2023-03-14
Rename 'publication' to 'reference' in the Publications table.

### LINKED ISSUE #208: Use consistent naming of Publications and Reference
It's confusing because the field name in every table is `reference` but it points to the Publications table. I propose we rename the Publications table to the References table.

kelle:
Change `publication` to `reference` in the Publications table so that it's more consistent with `reference` usage in all other tables.

Closes #208 

Tests aren't passing locally but I think it's because of sqlalchemy2.0 change. I may need to upgrade my local.


dr-rodriguez:
When you click the button to merge, consider the dropdown arrow beside it to see if squash and merge may be the better option here.

dr-rodriguez (changes requested):
Documentation update needed

dr-rodriguez (approved):

PR #352, opened 2023-03-15
MKO photometry - 2nd Try

SherelynA:
Changing old MKO photometry names to correct filter identifiers.

### DIFF
--- a/scripts/MKO_photometry_update.py
+++ b/scripts/MKO_photometry_update.py
@@ -2,7 +2,7 @@ from scripts.ingests.utils import *
 from astropy.table import Table
 from sqlalchemy import and_
 
-SAVE_DB = False  # save the data files in addition to modifying the .db file
+SAVE_DB = True  # save the data files in addition to modifying the .db file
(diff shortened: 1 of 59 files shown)


scripts/MKO_photometry_update.py
```diff
@@ -0,0 +1,103 @@
...
+SAVE_DB = True  # save the data files in addition to modifying the .db file
+RECREATE_DB = True  # recreates the .db file from the data files
+
+logger.setLevel(logging.INFO)
+
+db = load_simpledb('SIMPLE.db', recreatedb=RECREATE_DB)
+
+# Fix the Reid02.2806 Publication in the Photometry table and take out old MKO photometry filters
+if RECREATE_DB:
+    with db.engine.connect() as conn:
```
dr-rodriguez, line 14: Similar to the other comment, to make more effective use of this one you could have added `conn.commit` calls after each individual loop- that way each step is being committed as it works through the process rather than all at the end (which is what `.begin` does)

dr-rodriguez (approved):

PR #357, opened 2023-03-21
ingest filippazzo 2015 modeled parameters

Ajb2307:
Short description: Include what type of data being ingested and appropriate references.

Link to relevant issue: Closes #

For data ingests:

- [x]  includes script used for ingest

- [x] code to correctly format the table in the paper for modeled parameters

- [x]  Have checks to make sure all data is correctly named

- [x] separate existing vs. nonexisting sources


Ajb2307:
https://docs.google.com/spreadsheets/d/1BFKU0wAQsMmDqd24GsC4WrOgIUhmHna84B9zUz9EDmE/edit?usp=sharing

Google sheet of the data

scripts/ingests/filip_15.py
```diff
@@ -0,0 +1,10 @@
+from astrodbkit2.astrodb import create_database
+from astrodbkit2.astrodb import Database
+from astropy.table import Table
+
+connection_string = 'sqlite:///SIMPLE.db'  # connection string for a SQLite database named SIMPLE.db
+create_database(connection_string)
+db = Database(connection_string)
+db.load_database('data/')
```
kelle, line 8:
```suggestion
SAVE_DB = True  # save the data files in addition to modifying the .db file
RECREATE_DB = True  # recreates the .db file from the data files
VERBOSE = False
# LOAD THE DATABASE
db = load_simpledb('SIMPLE.db', recreatedb=RECREATE_DB)
```

kelle: And add this to the end:
```
# WRITE THE JSON FILES
if SAVE_DB:
    db.save_database(directory='data/')
```

scripts/ingests/table_9_filip_15.txt
Ajb2307, line 1: this is the original file from the paper

scripts/ingests/filip_15.py
```diff
@@ -0,0 +1,133 @@
...
+
+SAVE_DB = False  # save the data files in addition to modifying the .db file
+RECREATE_DB = True  # recreates the .db file from the data files
+VERBOSE = False
+# LOAD THE DATABASE
+db = load_simpledb('SIMPLE.db', recreatedb=RECREATE_DB)
+
+# WRITE THE JSON FILES
+if SAVE_DB:
+    db.save_database(directory='data/')
```
kelle, line 16: This should go at the end of your script.

scripts/ingests/filip_15.py
```diff
@@ -0,0 +1,133 @@
...
+    content = f.readlines()
+    table_9 = Table.read(content, format='ascii', data_start=4, data_end=201, header_start=2, delimiter='\t',
+                         guess=False, fast_reader=False, include_names=columns)
+
+data_columns = ['Radius', 'log(g)', 'T_eff', 'Mass']  # columns with data values
+
+for i in table_9:  # rewriting so that the empty values are None
+    for c in data_columns:
+        if i[c] == 'cdots':
+            i[c] = None
```
kelle, line 29:
```suggestion
for row in table_9:  # rewriting so that the empty values are None
    for column in data_columns:
        if row[column] == 'cdots':
            row[column] = None
```

scripts/ingests/filip_15.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def get_float_value(value):
+    # rewriting so that the string values are floats
+    if value == 'None':
+        return None
+    else:
+        float_value = float(value.split(' +or- ')[0])
+        return float_value
```
kelle, line 38: Modify to also return the uncertainty, the number after the `+or-`

scripts/ingests/filip_15.py
```diff
@@ -0,0 +1,133 @@
...
+fix_source_name(134, designations[133])
+fix_source_name(117, designations[118])
+fix_source_name(183, designations[182])
+fix_source_name(136, designations[135])
+fix_source_name(153, designations[152])
+fix_source_name(144, designations[143])
+fix_source_name(165, designations[164])
+fix_source_name(133, designations[184])
+fix_source_name(173, designations[172])
+fix_source_name(160, designations[159])
```
kelle, line 133: My suggestion is to all of this outside of Python and in Google Sheets or Excel, by hand.

kelle:
This is looking great!  Love all the functions that you're writing -- great to see modular code!

Major suggestion: Combine Table 1 and Table 9 by hand. Open them both up with Google Sheets or Excel or whatever and then copy and paste and do the designation-short name matching by hand.  Save that as a new table which you read into your script. The new table will have the designation, RA, Dec, AND model parameters all in one table.

That new table will be the one you loop through.

The next step will be to find the matching source in the database. We've written a the [`find_source_in_db`](https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/scripts/ingests/utils.py#L81) function to help with that.

Ajb2307:
https://docs.google.com/spreadsheets/d/1MOvy2s_W2vxUnoXu6uM7gIzg3azyZNpxigJCocDM-2o/edit?usp=sharing

new document

scripts/ingests/filip_15.py
```diff
@@ -14,16 +14,17 @@
...
+data9stuff = Table.read(link, format='ascii', data_start=6, data_end = 203, header_start=4, names=columns, guess=False,
+                        fast_reader=False, delimiter=',')
 
 data_columns = ['Radius', 'log(g)', 'T_eff', 'Mass']  # columns with data values
 
-for i in table_9:  # rewriting so that the empty values are None
+for i in data9stuff:  # rewriting so that the empty values are None
     for c in data_columns:
         if i[c] == 'cdots':
             i[c] = None
```
kelle:
```suggestion
for row in data9stuff:  # rewriting so that the empty values are None
    for column in data_columns:
        if row[column] == 'cdots':
            row[column] = None
```

Ajb2307: I forgot about this changing now

scripts/ingests/filip_15.py
```diff
@@ -0,0 +1,76 @@
...
+           'R.A.', 'Decl.', 'Designation 1', 'Discovery References', 'SpT', 'SpT References']
+filip15_table = Table.read(link, format='ascii', data_start=6, data_end=203, header_start=4, names=columns, guess=False,
+                           fast_reader=False, delimiter=',')
+
+data_columns = ['Radius', 'log(g)', 'T_eff', 'Mass']  # columns with wanted data values
+
+for row in filip15_table:  # rewriting so that the empty values are None
+    for column in data_columns:
+        if row[column] == 'cdots':
+            row[column] = None
```
kelle, line 31: rather than looping through each row, you can use the `np.where` function to find the "cdots". In general, you want to avoid loops,  where possible.
```suggestion
# replacing empty values ('cdots') with None
    for column in data_columns: 
        filip15_table[column][np.where(filip15_table[column] == 'cdots')] = None
```
The links that I used to help figure this out:
- https://docs.astropy.org/en/stable/table/modify_table.html
- https://stackoverflow.com/questions/16642022/numpy-where-on-string

scripts/ingests/filip_15.py
```diff
@@ -0,0 +1,76 @@
...
+        float_value = float(value.split(' +or- ')[0])
+        return float_value
+
+
+def get_coords(string):
+    # given RA or Dec as string given in table returns as degree decimals
+    angle = Angle(string, u.degree)
+    return angle.degree
+
+
```
kelle, line 56: You can use the map() function to perform the functions on the relevant columns:
```suggestion
for column in data_columns:
    filip15_table.add_column(list(map(get_float_value,filip15_table[column])), name=f"{column}_value")
    filip15_table.add_column(list(map(get_float_error,filip15_table[column])), name=f"{column}_error")

```

You'd need to write a new function to get the error.

scripts/ingests/filip_15.py
kelle: I also just realized you could do this in your `get_float_value` function.

kelle (changes requested):
The code looks great!!  Love the np.where and the way you handled the nested dictionaries. I also ran it locally and the data files look good.

Next thing to do is to take a look at the tests. Even though they are passing here, they are not passing locally on my laptop. Please take a look at the tests in `test_integrity` and now that there is data in the tables, you probably want to re-write them.

Also, since this is an ingest, also need to add test to `test_data`. Could be as simple as the number of mass, radius, logg, Teff values have Fili15 as a ref.

kelle (approved):
Looks great!!

PR #376, opened 2023-05-16
Adding Companion Relationships table

### LINKED ISSUE #230: Add Companion Relationships table.
When there are multiple objects in a system, it can get messy.  I propose that we have a separate source for each resolved object. E.g., Luhman 16AB would be three sources. 
- Luhman 16AB - the unresolved system
- Luhman 16A - the resolved primary
- Luhman 16B - the resolved secondary

If we do that, do we also want to add a column -- or a new table? -- which somehow connects the sources to each other? 

We could follow the SIMBAD model of using parent, children, and siblings. We could make a new dedicated table (Systems?), or add three columns to the Sources table for parent, children, and siblings.

kelle:
Working on adding the Companion Relationships table to the schema. Will also include a couple sources. Should also add some tests.

Link to relevant issue: Closes #230

### DIFF
--- a/documentation/CompanionRelationships.md
+++ b/documentation/CompanionRelationships.md
@@ -1,0 +1,22 @@
+# CompanionRelationships
+
+The CompanionRelationships table contains companions to sources listed in the Sources table. 
+The combination of *source* and *companion_name* is expected to be unique.
+Columns marked with an asterisk (*) may not be empty.
+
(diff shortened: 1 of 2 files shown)


kelle:
Todo:

- [ ] add docs page
- [ ] add `projected_separation` and `units` columns. 
- [ ] add tests

kelle:
Superseded by #385

PR #379, opened 2023-06-28
Rework photometry filters

### LINKED ISSUE #293: Tweak PhotometryFilters table
Right now, we require Telescope and Instrument in the PhotometryFilters Table but I think that's redudant since those fields should be associated with a particular observation. For example, lots of different instrument/telescope combos use the SDSS filter set.  I think we should remove those columns from the PhotometryFilters table and only require `band`.

dr-rodriguez:
This updates the Photometry and PhotometryFilters table in various ways:
 - UCD column moved from Photometry to PhotometryFilters
 - instrument and telescope columns removed from PhotometryFilters
 - Foreign Key of Photometry reworked to point to Telescopes
 - instrument column removed from Photometry (in anticipation of #380)

Closes #293


kelle:
This looks good but I admit to not having my head fully wrapped around it.  Should we write some tests to go along with it?

dr-rodriguez:
We have tests that query Photometry and PhotometryFilters already, I'm not sure what else we would add to them. The database will not build if the foreign keys and similar are not set.

I did notice the tests failing with the Instruments change (#380), but that's a separate PR.

kelle (approved):

PR #380, opened 2023-06-29
Rework Instruments table

### LINKED ISSUE #99: Instrumentation Uniqueness
Instrument should not be primary key but telescope-instrument combo; to allow for duplicated instrument names (e.g. nirspec/ osiris)

dr-rodriguez:
This reworks the Instruments table to be used for both mode and instrument name.
This is built on top of the ReworkPhotometryFilters branch which is why there are so many changes. Once #379 is merged this should be reviewed.

Closes #99


Will-Cooper (approved):
LGTM, doesn't look like any changes will break website. We discussed how `ingest_instruments` in `ingest_utils` may need splitting. Also how Gaia has multiple instruments (same may be true for other spacecraft). Website will have to swap logic from mode is None to mode == 'missing'.

PR #381, opened 2023-07-02
Add new template schema

arjunsavel:
Short description: I've been chatting with @kelle and @dr-rodriguez about including a template schema — that is, a version of the SIMPLE schema that other scientists can adapt to their use cases. So far, I've tossed out tables that others might not use, and I've included a `DataPointerTable` object that other tables can subclass. 

I believe my next step is to produce a DBeaver map of the tables, after which I'll make some test data, write some tests, etc. Happy to iterate on this, of course!

Link to relevant issue: Relevant to #99

For data ingests:
[ ] includes script used for ingest
[ ] includes modified JSON files
[ ] Add new tests
[ ] Update the Versions table

### DIFF
--- a/simple/template_schema.py
+++ b/simple/template_schema.py
@@ -1,0 +1,159 @@
+# Schema for the SIMPLE database
+
+from sqlalchemy import Column, Float, ForeignKey, Integer, String, Enum, DateTime, ForeignKeyConstraint
+import enum
+from astrodbkit2.astrodb import Base
+
+
+# -------------------------------------------------------------------------------------------------------------------
+# Reference tables
+class Publications(Base):
+    """ORM for publications table.
+    This stores reference information (DOI, bibcodes, etc) and has shortname as the primary key
+    """
+    __tablename__ = 'Publications'
+    reference = Column(String(30), primary_key=True, nullable=False)
+    bibcode = Column(String(100))
+    doi = Column(String(100))
+    description = Column(String(1000))
+
+
+class Telescopes(Base):
+    __tablename__ = 'Telescopes'
+    telescope = Column(String(30), primary_key=True, nullable=False)
+    description = Column(String(1000))
+    reference = Column(String(30), ForeignKey('Publications.reference', onupdate='cascade'))
+
+
+class Instruments(Base):
+    __tablename__ = 'Instruments'
+    instrument = Column(String(30), primary_key=True, nullable=False)
+    description = Column(String(1000))
+    reference = Column(String(30), ForeignKey('Publications.reference', onupdate='cascade'))
+
+# instead of having modes...telescope --> instrument --> mode (used mostly for spectra). think about SVO.
+class Modes(Base):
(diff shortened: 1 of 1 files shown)


dr-rodriguez:
As a note, #379 and #380 are reworking Modes/Instruments/Telescopes/PhotometryFilters and your PR may benefit from what is hopefully a simpler setup.

simple/template_schema.py
```diff
@@ -0,0 +1,159 @@
...
+class Names(Base):
+    __tablename__ = 'Names'
+    source = Column(String(100), ForeignKey('Sources.source', ondelete='cascade', onupdate='cascade'),
+                    nullable=False, primary_key=True)
+    other_name = Column(String(100), primary_key=True, nullable=False)
+
+# todo: make "tabulardata" or "physicaldata" abstract classes.
+
+
+class DataPointerTable(Base):
```
dr-rodriguez, line 113: You may want to see as an example what I implemented in https://github.com/dr-rodriguez/pycaomloader/blob/main/pycaomloader/schema.py#L39 This creates a class that does not inherit from `Base` and then is used in subsequent tables to share the same columns across them.

simple/template_schema.py
```diff
@@ -0,0 +1,159 @@
...
+    reference = Column(String(30), ForeignKey('Publications.publication', onupdate='cascade'), primary_key=True)
+    other_references = Column(String(100))
+    # __mapper_args__ = {'polymorphic_on': data_type}
+
+
+class Spectra(DataPointerTable):
+    # Table to store references to spectra
+    super().__init__()
+    __tablename__ = 'Spectra'
+    __mapper_args__ = {'polymorphic_identity': 'spectra'}
```
dr-rodriguez, line 138: I don't know what `polymorphic_identity` means; it may be a complexity you don't need to include in the template. If you DO need it, it may be worthwhile to add a link to a SQLAlchemy document that describes why.

simple/template_schema.py
```diff
@@ -0,0 +1,159 @@
...
+    """ORM for the sources table. This stores the main identifiers for our objects along with ra and dec"""
+    __tablename__ = 'Sources'
+    source = Column(String(100), primary_key=True, nullable=False)
+    ra = Column(Float)
+    dec = Column(Float)
+    epoch = Column(Float)  # decimal year
+    equinox = Column(String(10))  # eg, J2000
+    shortname = Column(String(30))  # not needed?
+    # reference = Column(String(30), ForeignKey('Publications.publication', onupdate='cascade'), nullable=False)
+    other_references = Column(String(100))
```
dr-rodriguez, line 100: Curious why you would remove reference in favor of other_reference. If all you wanted was to remove the ForeignKey relationship, it's easier to do just that than create a new column.

kelle:
This work is being implemented in a dedicated astro template repo: https://github.com/astrodbtoolkit/astrotemplate-db

PR #382, opened 2023-07-05
First Companion Table Ingest

Ajb2307:
First ingest into Companions table

- [x] ingest reference not in Simple
- [x] ingest source not in Simple
- [x] import google sheets
- [x] write ingest function


REFERENCE_TABLES.py
```diff
@@ -0,0 +1,8 @@
+from astrodbkit2.astrodb import create_database,Database
+from simple.schema import *
+
+connection_string = 'sqlite:///SIMPLE.db.py' # connection string for a SQLite database named SIMPLE.db.py
+create_database(connection_string)
+db = Database(connection_string)
+
+db.load_database('data/')
```
kelle, line 8: Not sure where you got this from. Please use your `filip_15.py` script as an example and use the `load_simpledb` function.

SIMPLE.db-journal
kelle, line 1: This file should not be committed.

SIMPLE.db.py
kelle, line 1: This file should not be committed.

REFERENCE_TABLES.py
Ajb2307: I deleted that file, honestly no clue where it came from

Ajb2307 (commented):
Add other names function I can only find in the  utils_deprecated file

kelle:
Looking good!!

Issue #360, opened 2023-03-22
Ingest Y-band photometry from the Ultracool Sheet


kelle:
The MKO Y photometry is the highest priority photometry that needs to be ingested from the Ultracool Sheet: https://docs.google.com/spreadsheets/d/1i98ft8g5mzPp2DNno0kcz4B9nzMxdpyz5UquAVhz-U8/edit#gid=1469954551

Also add ULASS names as appropriate.

(Separating this task from #81)

kelle:
I made a pseudo-script which outlines the possible necessary steps: https://gist.github.com/kelle/6ba958606ca820c096adf0e4ce0b3180

kelle:
Regarding the filter name: Not sure the MKO_Y is actually UKDISS.Y or UFTI.Y or something else. I think Ben B. will know.

PR #388, opened 2023-07-18
partial ingest codes

astro-c:
Y MKO data ingests

Update on #360

To do:
- [x] create lists for multiple instrumentation publications
- [x] finish manual ingest of publications
- [ ] pytest for ingest
- [x] pytest for photometry bands


Ajb2307:
Hi!

.gitignore
```diff
@@ -5,4 +5,5 @@
 .vscode
 .pytest_cache
 .ds_store
-bdnyc/
\ No newline at end of file
+bdnyc/
+/test.py
```
kelle: Tell me more about why you're editing the .gitignore.

kelle (commented):
This is looking good! 

- I would prefer the ingest script be just one file. Basically, incorporate MKO move utils functions into the ingest script. 

- Next step will be to add tests

.gitignore
astro-c: I was not meant to commit that change, sorry. It's a remote file I was using to test certain aspects of the code, I'll try to revert the commit ASAP.

astro-c:
I believe I reverted the accidental commit of the .gitignore file and successfully merged the MKO utils with the MKO ingest

astro-c:
Ingestion commit needs to be reversed. There were 2 things I had fixed with commits recently that I believed I had pushed but the push seems to have failed so the ingests are wrong.

kelle:
Hey M, are you going to be able to finish this up or should we take it over. It's sooooo close!  Let us know.  Thanks!

astro-c:
I can finish it but I need some tech support on this and I do not have much spare time available.

kelle:
Could you say more about the problems you're having? It's probably best to use Slack.

Also, since your script is in good shape, I recommend considering starting fresh. 
- Make a copy of your scripts 
- start a fresh branch from upstream:main.
- re run the scripts.

We can discuss in more detail
in Slack and/or setup some zoom time to work together.

Issue #347, opened 2023-03-09
Fix the Links of Spectra for Suarez22


SherelynA:
The links for the spectra for the Suarez22 ingest don't work and need to be fixed to have the correct ones.

kelle:
Please be more specific about what "doesn't work" mean and give a couple examples.

SherelynA:
The issue lies in that the sources which have a plus in the name like 2MASS J21481633+4003594 have the url name of **https://bdnyc.s3.amazonaws.com/IRS/2148%252B4003_IRS_spectrum.fits** which does not work because the actual working link is **https://bdnyc.s3.amazonaws.com/IRS/2MASS+J21481633%2B4003594.fits**. This is the google doc link to the correct spectra urls :[https://docs.google.com/spreadsheets/d/1ojsNF59GvdFhf-lgwJaCwiHPfaRx4fmCpyzKEonrVbw/edit?usp=sharing](url)

kelle:
another example: https://simple-bd-archive.org/load_solo/2MASS%20J12130336-0432437

kelle:
There are 149 URLs giving 404s:
[...]

PR #404, opened 2023-09-19
add test for spectrum urls

### LINKED ISSUE #369: Make test which makes sure all the spectra URLs are valid
Similar code is in [`ingest_spectra#L1089`](https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/scripts/ingests/ingest_utils.py#L1089) but it should also go in the [`test_spectra`](https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/tests/test_integrity.py#L530) function of `test_integrity`.

kelle:
New test to make sure spectrum urls don't give 404.  This test takes some time....not sure we want to keep it, but it's useful at the moment. Related to #347. 

Link to relevant issue: Closes #369


dr-rodriguez:
My worry is that this may be attempting to fetch thousands of spectra for each time the tests runs, which is why it takes a longer time.   
Perhaps a better approach is that this test exists, but is in a separate file and there is a separate github action for a periodic (once per week maybe) execution as opposed to with every commit.

kelle:
YES!  Here's how to do it: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

What should we call the file? still test_<something>? If we do that, it will be run by default by pytest. So maybe a different name?

dr-rodriguez:
We could change the pytest defaults, or just call it something like scheduled_jobs.py or something similar

kelle:
scheduled_tests?

dr-rodriguez:
I don't know if the pytest defaults look at anything that starts with `test_` or just have `test` somewhere in it. We probably should explicitly have a pytest.ini file to control that. There's also a way to set markers on pytest so we could have slow/fast tests and configured the current tests to use the fast tests and the scheduled ones to use the slow ones. 
But if you want to be safe and not add any of that, maybe just avoid test in the filename.

kelle:
Here's the documentation for pytest autodiscovery:
https://docs.pytest.org/en/6.2.x/goodpractices.html#conventions-for-python-test-discovery

My first choice is to make a file/test which just doesn't get auto-discovered but still lives in `tests/`

tests/scheduled_checks.py
```diff
@@ -0,0 +1,57 @@
...
+
+    return db
+
+
+def test_spectra_urls(db):
+    spectra_urls = db.query(db.Spectra.c.spectrum).astropy()
+    broken_urls = []
+    codes = []
+    internet = check_internet_connection()
+    if internet:
```
dr-rodriguez, line 48: I'm not familiar with `check_internet_connection` but if this returns False, the if statement doesn't run but the assert does and so this test would fail.

tests/test_data.py
```diff
@@ -3,12 +3,14 @@
...
 import pytest
 import sys
+import requests
 
 sys.path.append('.')
 from simple.schema import *
 from astrodbkit2.astrodb import create_database, Database
 from sqlalchemy import except_, select, and_
 from . import REFERENCE_TABLES
+from scripts.ingests.utils import check_internet_connection
```
dr-rodriguez, line 13: Do you still need this import now that you moved the logic to scheduled_checks.py?

scripts/updates/fix_IRS_spectra_links.py
```diff
@@ -0,0 +1,26 @@
...
+
+SAVE_DB = True  # save the data files in addition to modifying the .db file
+RECREATE_DB = True  # recreates the .db file from the data files
+
+# LOAD THE DATABASE
+db = load_simpledb('SIMPLE.db', recreatedb=RECREATE_DB)
+
+
+#  link to live google sheet
+link = 'https://docs.google.com/spreadsheets/d/e/2PACX-1vQG5cGkI2aHPHD4b6ZZPTU4jjQMirU_z-yhl5ElI3p6nCIufIL64crC-yFalF58OauWHxmYvEKR_isY/pub?gid=0&single=true&output=csv'
```
dr-rodriguez, line 13: Any concern with explicitly having this link in the code base? Not sure if this is a public or private document, but it's now part of the version history for anyone to access.

kelle: I don't have any concern with this document being publicly discoverable.

tests/scheduled_checks.py
```diff
@@ -0,0 +1,57 @@
...
+        for spectrum_url in spectra_urls['spectrum']:
+            request_response = requests.head(spectrum_url)
+            status_code = request_response.status_code  
+            # The website is up if the status code is 200
+            # cuny academic commons links give 301 status code
+            if status_code != 200 and status_code != 301:
+                broken_urls.append(spectrum_url)
+                codes.append(status_code)
+    assert len(broken_urls) == 149, f'found {len(broken_urls)} broken spectra urls: {broken_urls}, {codes}'
```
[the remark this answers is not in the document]
kelle: I actually like the tests to be `==`. If the number changes in a way which we didn't expect, then I want to know abuot it.

kelle:
This is ready for re-review.  It's *just* the tests. Once it's merged, I'd like to run the action, make sure it works with all the broken URLs.

Then I can work on #409 to actually fix the URLs.

PR #407, opened 2023-09-26
Parametrizing the photometry tests for clearer updates

### LINKED ISSUE #406: Rework test_data to parametrize testing photometry
Use pytest.mark.parametrize and iterate over each. This will be better for long term testing as each photometry band can be tested independently (and will compact some of the re-used code)

dr-rodriguez:
Rework one of the tests for easier updates. We can probably apply this strategy to other tests, but for now just on photometry.

Closes #406


Will-Cooper (approved):
LGTM, let's pin this as the optimal way of testing lots of the same thing

PR #411, opened 2023-10-17
Fix type scheduled-tests.yml

kelle:
Trying to get the scheduled tests to work!


dr-rodriguez (approved):

kelle:
- added `tqdm` progress bar. Very satisfying.
- Number failed alternated between 149 and 150 so made assert a range.

PR #427, opened 2023-12-05
First ingest wise_1810-1010

### LINKED ISSUE #258: Ingest Wise 1810-1010, metal poor T dwarf
Discovered by Schneider at Al 2020

image

LishaRamon:
Data of Wise_1810-1010 that are ingested in this script are publications, sources, names and parallaxes.

Link to relevant issue: Closes #258 

For data ingests:
- [x] Add new tests


LishaRamon:
Will be adding tests next.

kelle:
This looks good but somehow you're also modifying `2mass_j12560183-1257276.json`.  That should not be part of this pull request.

LishaRamon:
A .json file wasnt produced when saving the db, assuming because the code doesn't ingest yet. There is an error with source not being recognized as well.

kelle (changes requested):

LishaRamon (commented):
Viewed and implemented changes

LishaRamon:
Trying to push current code, but saying I need to do a pull request first. Approved the changes you sent and implemented them. Used git fetch to make sure the files are the same. Not able to push at the moment, any suggestions?

LishaRamon:
I believe all of them pushed (showed up together). Most recent one (push attempt) should be the current one. One issue is there being a similar publication in the data. And occasionally says "add_publications" is not defined.

data/2mass_j12560183-1257276.json
kelle, line 1: This file shouldn't be part of the ingest. My advice is to modify the file by hand to make it look exactly the way it is on upstream:main.

kelle (changes requested):

LishaRamon:
Both commits have the same script, just got pushed twice on accident.

scripts/ingests/ingest_wise_1810-1010.py
```diff
@@ -0,0 +1,108 @@
...
+                                        'parameter': "mass", 
+                                        'unit': 'M_jup',
+                                        'reference': "Lodi22"}
+                                     }]
+                                   
+    source = "CWISEP J181006.00-101001.1"
+    value_types = ['Gravity', 'Metallicity', 'Radius', 'Mass']
+    with db.engine.connect() as conn:
+        for row in ingest_modeled_parameters:
+            row_dict = []
```
kelle, line 92: this variable isn't used, I think it can be deleted.

scripts/ingests/ingest_wise_1810-1010.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+#Ingest Functions for Modeled Parameters (Gravity, Metallicity, Radius, Mass)
+#Creating list of dictionaries for each value formatted for modeled parameters
+def add_modeled_parameters_dict(db):
+    ingest_modeled_parameters = [{ 
+                                    'Gravity':
+                                       {'value': 5.0,
+                                        'value_error': 0.25,
+                                        'parameter': "Radius",
```
kelle, line 64: this should be 'log g', not 'Radius'.

scripts/ingests/ingest_wise_1810-1010.py
```diff
@@ -0,0 +1,108 @@
...
+                                   'Metallicity':
+                                       {'value': -1.5,
+                                        'value_error': 0.5,
+                                        'parameter': "Metallicity", 
+                                        'unit': 'dex',
+                                        'reference': "Lodi22"},
+                                   'Radius':
+                                       {'value': 0.067,
+                                        'value_error': 0.032, #Highest value error was picked between +0.032 & -0.020 listed
+                                        'parameter': "Radius",
```
kelle, line 77:
```suggestion
                                        'parameter': "radius", 
```

kelle:
The parameter names MUST be in the `Parameters` table.  That's why you're getting foreign key constraint errors. "Radius" is not in the `Parameters` table.

scripts/ingests/ingest_wise_1810-1010.py
LishaRamon: is log_g another name for radius?

kelle: This is the dictionary for the Gravity parameter.  `log g` is another name for gravity. It's the log of the gravity.

LishaRamon (commented):
I thought since we were ingesting modeled parameters, that meant we would be adding other types of parameters(as in the new spelling/name would not matter). But I'm assuming since gravity/radius were ingested by other data before, that means we need to follow the same parameter spelling as what was previously ingested? Meaning new parameters like metallicity will have to be ingested as 'metallicity' from now on?

The new changes were implemented in the json file and the parameters are reflected in the db!

kelle:
Good work!

- The parameters added to the `ModeledParameters` table must be named and described in the `Parameters` table.  Take a look at that table in the db browser and read the documentation carefully. https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/documentation/ModeledParameters.md

- Also add the effective temperature. The value is in that table linked from the issue.
- Make your 1256 file look exactly like the one [here](https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/data/2mass_j12560215-1257217.json)

kelle:
Yes, you're right. I linked to the wrong file.

kelle:
And yes, effective temperature is another parameter.

LishaRamon:
Since the value is less than or equal to 1000, would there be an appropriate error value to input? 
How would we reference the poster in the db?

scripts/ingests/ingest_wise_1810-1010.py
```diff
@@ -0,0 +1,115 @@
...
+                                   'Mass':
+                                       {'value': 17,
+                                        'value_error': 56, #Highest value error was picked between +56 & -12 listed
+                                        'parameter': "mass", 
+                                        'unit': 'M_jup',
+                                        'reference': "Lodi22"},
+
+                                    'Effective Temperature':
+                                       {'value': 1000,
+                                        'parameter': "temperature",
```
kelle, line 91: you have "temperature".  Look at the Parameters table to see what the parameter name should be.

kelle: also add the value_error.

kelle:
Look at Table 6 from the paper that's linked in the issue. Do not use the poster values.

data/cwisep_j181006.00-101001.1.json
```diff
@@ -0,0 +1,77 @@
...
+            "unit": "dex",
+            "comments": null,
+            "reference": "Lodi22"
+        },
+        {
+            "parameter": "mass",
+            "value": 17.0,
+            "value_error": 56.0,
+            "unit": "M_jup",
+            "comments": null,
```
kelle, line 29: Let's add the asymmetric uncertainty in the comment.  
```suggestion
            "comments":"12[+56, -12]"
```

data/cwisep_j181006.00-101001.1.json
```diff
@@ -0,0 +1,77 @@
...
+            "value_error": 0.5,
+            "unit": "dex",
+            "comments": null,
+            "reference": "Lodi22"
+        },
+        {
+            "parameter": "radius",
+            "value": 0.067,
+            "value_error": 0.032,
+            "unit": "R_jup",
```
kelle, line 44: The value in the paper is actually in solar radaii, not jupiter radii.  You need to convert. I recommend using astropy units. https://docs.astropy.org/en/stable/constants/.  https://docs.astropy.org/en/stable/units/quantity.html#quantity-unit-conversion

LishaRamon: 0.067 and 0.032 needs to be in solar radii, which need to be coded into the json file with astropy?

kelle: Try just changing the unit.  R_sun instead of R_jup.

data/cwisep_j181006.00-101001.1.json
kelle: should be 17.

kelle: Let's also make error null.

data/2mass_j12560183-1257276.json
```diff
@@ -1,4 +1,4 @@
-{
+{
```
kelle, line 1:
```suggestion
{
```

LishaRamon:
The commit named "Jsons updated / t eff edit needed" with the number '[b6e5a08]" is the correct one with current changes. Others were repeated trial push attempts with the previous error. Just to make 4 file updates easier to see:

Versions.py- latest version was added/updated
2MASS J12560183-1257276.json- checked it was identical to the original db
CWISEP J181006.00-101001.1.json- under modeled parameters: mass(value_error changed to null and comments added/corrected), radius(unit changed to R_sun)
ingest_wise_1810-1010.py- Effective Temp parameters were added, checked, still does not show in db/json file

tests/test_data.py
```diff
@@ -557,6 +557,16 @@ def test_modeledparameters(db):
     t = db.query(db.ModeledParameters).filter(db.ModeledParameters.c.reference == ref).astropy()
     assert len(t) == 696, f'found {len(t)} modeled parameters with {ref} reference'
 
+    #Test to verify T eff counts
+    ref = 'T eff'
```
kelle, line 561:
```suggestion
    parameter = 'T eff'
```
Let's rename this variable just to be more accurate.

tests/test_data.py
```diff
@@ -557,6 +557,16 @@ def test_modeledparameters(db):
...
+    t = db.query(db.ModeledParameters).filter(db.ModeledParameters.c.parameter == ref).astropy()
+    assert len(t) == 174, f'found {len(t)} modeled parameters with {ref} parameter'
```
kelle, line 563:
```suggestion
    param = 'T eff'
    t = db.query(db.ModeledParameters).filter(db.ModeledParameters.c.parameter == param).astropy()
    assert len(t) == 174, f'found {len(t)} modeled parameters with {param} parameter'
```
let's rename `ref` to be `param` so it's more readable/accurate.

kelle:
Good job getting the push to work!

I pushed a commit which removed the 1256 json file so you'll need to pull again.

did you save the database after adding the T_eff?

kelle:
T eff test looks great!  Go ahead and add tests for the other parameters.

kelle (approved):
looks great!!!

Issue #413, opened 2023-10-17
Fix broken 4 spectra links


kelle:
There are 30 spectra which are giving 404 errors.

[...]

PR #449, opened 2024-01-03
Fixed broken spectra links

kelle:
Progress on broken spectra links. #413 

Changes number expected from 39 to 4. Woot!


kelle:
Confirmed that the `scheduled_checks` scripts runs successfully.

LishaRamon (approved):
All good!

PR #451, opened 2024-01-08
Script to generate URLs for Ultracool Sheet. Closes #389

### LINKED ISSUE #389: Column Linking Objects to Ultracool Sheet
On discussion with Trent, it would be good to be able to link from SIMPLE directly to the [ultracool sheet](https://docs.google.com/spreadsheets/d/1i98ft8g5mzPp2DNno0kcz4B9nzMxdpyz5UquAVhz-U8/edit#gid=1469954551) and vice versa, to get people using SIMPLE.

kelle:
Here is a script to generate the SIMPLE URL for sources in the Ultracool Sheet.

This draft tries to do URL checking but unfortunately, any invalid URLs give status code 200. We need to figure out how to do proper URL checking.

### DIFF
--- a/scripts/ultracool_sheet/generate_simple_links.py
+++ b/scripts/ultracool_sheet/generate_simple_links.py
@@ -1,0 +1,81 @@
+#  from scripts.ingests.ingest_utils import *
+from scripts.ingests.utils import load_simpledb, find_source_in_db
+from astropy.io import ascii
+import urllib.request
+from astropy.table import Table
+
+
+RECREATE_DB = False
+db = load_simpledb("SIMPLE.sqlite", recreatedb=RECREATE_DB)
+
+# Load Ultracool sheet
+sheet_id = "1i98ft8g5mzPp2DNno0kcz4B9nzMxdpyz5UquAVhz-U8"
+link = f"https://docs.google.com/spreadsheets/d/{sheet_id}/export?format=csv"
+link = "scripts/ultracool_sheet/UltracoolSheet - Main_010824.csv"
+
+# read the csv data into an astropy table
+uc_sheet_table = ascii.read(
+    link,
+    format="csv",
+    data_start=1,
+    header_start=0,
+    guess=False,
+    fast_reader=False,
+    delimiter=",",
+)
+
+# Match sources in Ultracool sheet to sources in SIMPLE
+uc_names = []
+simple_urls = []
+simple_sources = []
+for source in uc_sheet_table[0:100]:
+    uc_sheet_name = source["name"]
+    match = find_source_in_db(
+        db,
+        uc_sheet_name,
+        ra=source["ra_j2000_formula"],
+        dec=source["dec_j2000_formula"],
+    )
+
+    # convert SIMPLE source name to URL
+    if len(match) == 0:
+        print("No match found for ", uc_sheet_name)
+        raise ValueError
+    elif len(match) > 1:
+        print("Multiple matches found for ", uc_sheet_name)
+        raise ValueError
+    elif len(match) == 1:
+        simple_source = match[0]
+        print(f"Match found for {uc_sheet_name}: {simple_source}")
+    else:
+        raise ValueError("Unexpected state")
+
+    # URLify source name
+    source_url = simple_source.strip().replace(" ", "%21")
+    url = "https://simple-bd-archive.org/load_solo/" + source_url
+    # u rl = "https://simple-bd-archive.org/solo_result/" + source_url
+
+    # TODO: THIS DOESN'T WORK!!! Even bad URLs return 200
+    # test the URL
+    url_status = 200
+    # url_status = urllib.request.urlopen(url).getcode()
+    if url_status != 200:
+        raise ValueError("URL not valid for ", uc_sheet_name, simple_source, url)
+    else:
+        print("URL valid for ", uc_sheet_name, simple_source, url)
+
+    uc_names.append(uc_sheet_name)
+    simple_sources.append(simple_source)
+    simple_urls.append(url)
+
+# write the results to a file
+results_table = Table(
+    [uc_names, simple_sources, simple_urls],
+    names=["Ultracool Sheet Name", "SIMPLE Source Name", "SIMPLE URL"],
+)
+results_table.write(
+    "scripts/ultracool_sheet/uc_sheet_simple_urls.csv",
+    delimiter=",",
+    overwrite=True,
+    format="ascii.ecsv",
+)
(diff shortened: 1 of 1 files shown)


scripts/ultracool_sheet/generate_simple_links.py
```diff
@@ -0,0 +1,81 @@
...
+        raise ValueError(msg)
+    elif len(match) == 1:
+        simple_source = match[0]
+        print(f"Match found for {uc_sheet_name}: {simple_source}")
+    else:
+        raise ValueError("Unexpected state")
+
+    # URLify source name
+    source_url = simple_source.strip().replace(" ", "%21")
+    url = "https://simple-bd-archive.org/load_solo/" + source_url
```
Will-Cooper, line 55:
```
from urllib.parse import quote
```
```suggestion
    source_url = quote(simple_source)
    url = "https://simple-bd-archive.org/load_solo/" + source_url
```

Will-Cooper (changes requested):
can use `urllib` to handle strings

kelle:
Thanks for the suggestion!

Any thoughts on how to properly test the links? Right now, even bad links return 200 because you're doing good error handling on the website.

Will-Cooper:
Hmm, how about 
```
results: dict = db.inventory(simple_source)
assert('Sources' in results.keys())
```
or that sort of logic?
Because anything that passes that will have a website page. Or are we trying to specifically ascertain that the created url will be correct?

Will-Cooper:
Alternatively, I could edit the logic on the website to fail on a bad result and return a 404.

Will-Cooper:
Okie dokie I'll see what I can do

Will-Cooper:
Returning the exact text as would appear on the 404 page on the website isn't quite as simple and not really worthwhile (would have to make custom headers etc), because of how WSGI handles things server side. It's not really the done thing to override the default status code message.
But since that PR on the website, it will return a 404 code to an API request instead of 200.

scripts/ultracool_sheet/generate_simple_links.py
```diff
@@ -0,0 +1,81 @@
+#  from scripts.ingests.ingest_utils import *
+from scripts.ingests.utils import load_simpledb, find_source_in_db
+from astropy.io import ascii
+import urllib.request
```
Will-Cooper, line 4:
```suggestion
from urllib.parse import quote
import requests
```

scripts/ultracool_sheet/generate_simple_links.py
```diff
@@ -0,0 +1,81 @@
...
+    # u rl = "https://simple-bd-archive.org/solo_result/" + source_url
+
+    # TODO: THIS DOESN'T WORK!!! Even bad URLs return 200
+    # test the URL
+    url_status = 200
+    # url_status = urllib.request.urlopen(url).getcode()
+    if url_status != 200:
+        raise ValueError("URL not valid for ", uc_sheet_name, simple_source, url)
+    else:
+        print("URL valid for ", uc_sheet_name, simple_source, url)
```
Will-Cooper, line 65: using `requests` instead of `urllib.request` because the former won't return an error if the request is 404
```suggestion
    url_status = requests.get(url).status_code
    if url_status != 200:
        raise ValueError("URL not valid for ", uc_sheet_name, simple_source, url)
    else:
        print("URL valid for ", uc_sheet_name, simple_source, url)
```

Will-Cooper (changes requested):

scripts/ultracool_sheet/generate_simple_links.py
```diff
@@ -0,0 +1,81 @@
...
+        raise ValueError("Unexpected state")
+
+    # URLify source name
+    source_url = quote(simple_source)
+    url = "https://simple-bd-archive.org/solo_result/" + source_url
+
+    # TODO: THIS DOESN'T WORK!!! Even bad URLs return 200
+    # test the URL
+    # url_status = 200
+    url_status = requests.get(url).status_code
```
Will-Cooper, line 61: Change to HEAD request instead of GET, to not overload the server and fail. After this change, this worked for me in the codespace.
```suggestion
    url_status = requests.head(url).status_code
```

Will-Cooper (changes requested):

scripts/ultracool_sheet/generate_simple_links.py
```diff
@@ -0,0 +1,80 @@
...
+    # URLify source name
+    source_url = quote(simple_source)
+    url = "https://simple-bd-archive.org/solo_result/" + source_url
+
+    # test the URL to make sure it is valid
+    url_status = requests.head(url).status_code
+    if url_status != 200:
+        raise ValueError("URL not valid for ", uc_sheet_name, simple_source, url)
+    else:
+        print("URL valid for ", uc_sheet_name, simple_source, url)
```
Will-Cooper, line 62:
```suggestion
    if url_status == 404:
        raise ValueError("URL not valid for ", uc_sheet_name, simple_source, url)
    elif url_status != 200:
        print("URL valid for ", uc_sheet_name, simple_source, url, " BUT the webpage is having issue: ", url_status)
    else:
        print("URL valid for ", uc_sheet_name, simple_source, url)
```

Will-Cooper (changes requested):

scripts/ultracool_sheet/generate_simple_links.py
kelle: Let's do something like give a warning if it's a timeout.

Will-Cooper, line 62:
```suggestion
    if len(match) == 0:
        msg = f"No match found for {uc_sheet_name}"
        logger.error(msg)
        raise SimpleError(msg)
    elif len(match) > 1:
        msg = f"Multiple matches found for {uc_sheet_name}"
        logger.error(msg)
        raise SimpleError(msg)
    elif len(match) == 1:
        simple_source = match[0]
        logger.info(f"Match found for {uc_sheet_name}: {simple_source}")
    else:
        msg = "Unexpected state"
        logger.error(msg)
        raise SimpleError(msg)

    # URLify source name
    source_url = quote(simple_source)
    url = "https://simple-bd-archive.org/solo_result/" + source_url

    # test the URL to make sure it is valid
    url_status = requests.head(url).status_code
    if url_status == 404:
        logger.error(f"URL not valid for {uc_sheet_name} {simple_source} at {url}")
        raise SimpleError("URL not valid for ", uc_sheet_name, simple_source, url)
    elif url_status != 200:
        logger.warning(f"URL valid for {uc_sheet_name} {simple_source} at {url} but with HTTP status {url_status}")
    else:
        logger.info(f"URL valid for {uc_sheet_name} {simple_source} at {url}")
```

scripts/ultracool_sheet/generate_simple_links.py
```diff
@@ -0,0 +1,80 @@
+from scripts.ingests.utils import load_simpledb, find_source_in_db
```
Will-Cooper, line 1:
```suggestion
from scripts.ingests.utils import load_simpledb, find_source_in_db, logger, SimpleError
```

Will-Cooper (changes requested):

Will-Cooper (approved):
LGTM, and I see from the email thread that you got a working csv of all matches!

Issue #372, opened 2023-05-16
Document how to use this repo


kelle:
We don't actually have that many dependencies.  Maybe we change our install instructions to just have them install `astrodbkit2` and maybe also `ads`.

Maybe we want to do the packaging work to make `pip install simple`.

kelle:
We could use setuptools. Here are some resources
- https://setuptools.pypa.io/en/latest/userguide/quickstart.html
- https://packaging-guide.openastronomy.org/en/latest/index.html

kelle:
Several questions:
- do we want data to get installed automatically with a `pip install`?
- where should the `utils` live? Should probably be someplace different than the ingest scripts.

kelle:
I _think_ the vision is for folks to **clone** this repo. 

Pip install AstroDB_scripts, but clone the database and custom scripts.

kelle:
This needs to be added to the README.

PR #457, opened 2024-01-24
Reorganize scripts into new package

### LINKED ISSUE #456: Rename `simple` folder to `schema`
To align with the astrodb-template structure, rename the `simple` folder to `schema`

kelle:
- scripts moved to new package
- schema folder renamed `schema`

Link to relevant issue: Closes #456 and #372


Will-Cooper:
minor thing to update import group order to be external libraries > standard libraries > local libraries, so `plotting_example` would become
```python
from astrodbkit2.spectra import load_spectrum
import astropy.units as u
import matplotlib.pyplot as plt
import numpy as np
from sqlalchemy import and_, cast, func, Integer

import logging

from scripts.ingests.ingest_utils import convert_spt_code_to_string_to_code
from scripts.ingests.utils import load_simpledb, logger
from scripts.ingests.simple_utils import convert_spt_code_to_string_to_code
```

kelle:
I appreciate this comment, but I don't think it's high priority to do that type of cleaning on the example scripts in this repo. The examples that end up in the template repo, on the other hand, should be as perfect as possible.

dr-rodriguez (approved):

PR #471, opened 2024-02-26
Modify the schema with a new Regimes table

kelle:
- Add a new Regimes table.  Closes #271 
- Make SpectralTypes, Spectra, and Gravities use Regimes.regime as a foreign key. CLoses #271 
- Implement conftest.py to create the database fixtures used by the tests. Closes #467.
- Moved REFERENCE_TABLE to  `schema.py` Closes #460 

~~Waiting for #470~~ - Done

PR #475, opened 2024-02-28
Add ingest_radial_velocity function. 

kelle:
Utility function and test to ingest radial velocities.


Will-Cooper (dismissed):
Could we please get some docstrings and type hints in the `ingest_gaiadr3` script? Is it true that whatever is the most recent upload is what we are taking as *adopted*? Can we also get a test of NaN or inf data, hopefully those should be rejected.

kelle:
The `ingest_gaiadr3` script was just copied and pasted from a stale PR.  It's in the `scripts\` directory, not the utils directory. I think it can receive more time and attention once it's actually used as part of an ingest. But right now, I just didn't want to lose the work.  So, long story short, I don't want to spend time tidying that script.

The function worthy of more scrutiny is the `ingest_radial_velocity` function in the `astrometry` module.

kelle:
I will add type hints to the ingest_radial_velocity script and something about units. 

Also, since there is not yet *any* rv data in SIMPLE, I will also modify the schema to add units to the column names.

kelle:
Because I've improved the testing setup in #471, I should really wait for it to be merged before progressing with this one.

.github/workflows/gen-db.yml
```diff
@@ -9,7 +9,6 @@ on:
   workflow_dispatch:  # manual execution
   release:
     types: [published]
-    branches: [main]
```
dr-rodriguez, line 12: Curious why this change? I guess we sometimes want to release from non-main branches?

simple/utils/astrometry.py
```diff
@@ -357,3 +357,181 @@ def ingest_proper_motions(
...
     return
+
+
+def ingest_radial_velocity(
+    db: Database,
+    *,
+    source: str = None,
+    rv: Union[Quantity, float] = None,
+    rv_err: Optional[Union[float, Quantity]] = None,
+    reference: str = None,
```
dr-rodriguez, line 368: I think the types here (and the docstring descriptions) are incorrect: these are not strings/floats/Quantities, but lists of string/floats/Quantities (or numpy arrays/astropy Tables or similar)

dr-rodriguez: Actually never mind, I think I confused `ingest_radial_velocity` with `ingest_radial_velocities`

dr-rodriguez: I find it interesting (and confusing) that this function takes a single value while the others in this module take lists of values- we should try to be consistent if possible.

scripts/ingests/Gaia/ingest_gaiadr3.py
```diff
@@ -0,0 +1,89 @@
+from scripts.ingests.ingest_utils import *
+from scripts.ingests.utils import *
```
dr-rodriguez, line 2: May be good to explicitly list which functions we are using from here rather than importing *

scripts/ingests/Gaia/ingest_gaiadr3.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+# update_ref_tables()
+
+
+def add_gaia_rvs(data, ref):
+    unmasked_rvs = np.logical_not(data["radial_velocity"].mask).nonzero()
+    rvs = data[unmasked_rvs]["db_names", "radial_velocity", "radial_velocity_error"]
+    refs = [ref] * len(rvs)
+    ingest_radial_velocities(
```
dr-rodriguez, line 63: This tripped me up because astrometry.py is in this PR and refers to a different singular function. Should we aim to rename these as `ingest_multiple_radial_velocities` or similar?

simple/utils/astrometry.py
```diff
@@ -357,3 +357,181 @@ def ingest_proper_motions(
...
+    -------
+    flags: dict
+        'added' : bool
+        'skipped' : bool
+
+    Examples
+    ----------
+    > ingest_radial_velocity(db, my_source, rv=my_rv, rv_err=my_rv_unc,
+                             reference=my_rv_ref,
+                             raise_error = False)
```
dr-rodriguez, line 400: Example may be better with some actual numbers/strings as opposed to variable names to demonstrate the types of inputs (similar to what's in the tests).

simple/utils/astrometry.py
Will-Cooper: I still hold that all functions in utils should be built to be single entry ingests but with like a `np.vectorize` decorator

dr-rodriguez (approved):

PR #479, opened 2024-03-05
fix generate_database action

kelle:
- Re-adds the `generate_database.py` script to generate the database file for the SIMPLE-binary repo.

- Changed database name from `.db` to `.sqlite`.


kelle:
Getting a username/password error. Not sure if that's because this is a pull requests or not.

dr-rodriguez:
> Getting a username/password error. Not sure if that's because this is a pull requests or not.

Likely yes, I think there are secrets either in this repo or in the binary repo that control how we can push to the binary repo. I don't know if pull requests can access this information.

simple/utils/generate_database.py
```diff
@@ -0,0 +1,45 @@
...
+    #    help="Connection string to use for non-sqlite databases.",
+    # )
+
+    args = parser.parse_args()
+
+    # Get the connection string for any non-sqlite database
+    # if args.connection_string is not None:
+    #    connection_string = args.connection_string
+    # else:
+    connection_string = os.getenv("SIMPLE_DATABASE_URL", default="")
```
dr-rodriguez, line 37: This doesn't look like it's being used.

.github/workflows/gen-db.yml
```diff
@@ -5,11 +5,12 @@
 name: Generate Database
 
 on:
-  # More details on trigger events: https://docs.github.com/en/actions/reference/events-that-trigger-workflows
+  # More details on trigger events: 
+  #  https://docs.github.com/en/actions/reference/events-that-trigger-workflows
+  pull_request:  # added for testing
```
dr-rodriguez, line 10: We should not do this on a pull_request, as this action essentially builds the binary file for purposes of the website. That should only be done when we are ready to push a change, not when someone opens a pull request.

.github/workflows/gen-db.yml
```diff
@@ -34,26 +35,16 @@ jobs:
...
-#      working-directory: .
-
     - name: Push database file
       uses: dmnemec/copy_file_to_another_repo_action@main
       # Details for this action at https://github.com/marketplace/actions/push-a-file-to-another-repository
       env:
           API_TOKEN_GITHUB: ${{ secrets.SIMPLE_TOKEN }}
       with:
-          source_file: 'SIMPLE.db'
+          source_file: 'SIMPLE.sqlite'
```
dr-rodriguez, line 47: Once this is merged, we will need to go to the binary repo and delete the old db file to avoid confusion.

dr-rodriguez:
Recommend avoiding `load_astrodb` for now, as that re-defines reference tables and such. It's best to have lower-level functions here (ie, direct calls to astrodbkit2) rather than this wrapper.

kelle:
> Recommend avoiding `load_astrodb` for now, as that re-defines reference tables and such. It's best to have lower-level functions here (ie, direct calls to astrodbkit2) rather than this wrapper.

I disagree.  This is exactly theh use case for `load_astrodb` and I'd rather just fix it at the source.  This PR is back in Draft mode while I make the needed changes in `astrodb_utils`.

kelle:
Waiting on https://github.com/astrodbtoolkit/astrodb_utils/pull/42

.github/workflows/gen-db.yml
kelle: I am pretty sure I added this here just so it would run on *this* PR. in order to test the script.

dr-rodriguez (approved):

PR #492, opened 2024-04-17
Photfilters update

### LINKED ISSUE #301: Photometry improvements
- add more UCDs to the Photometry table, 
- figure out the actual filters used for the Johnsons-Cousins.UBVRI and MKO.YJHK measurements.
- make a test which checks if filters are in the SVO format using their API: http://svo2.cab.inta-csic.es/svo/theory/fps3/index.php?mode=voservice

kelle:
Clean up the UCDs in the PhotometryFilters table.

Closes #301

For data ingests:
- [x] Add new tests


scripts/updates/update_phot_filters.py
```diff
@@ -0,0 +1,80 @@
...
+# Run a query to find the bands with null UCDs
+t = db.query(
+    db.PhotometryFilters.c.band,
+    db.PhotometryFilters.c.ucd,
+    db.PhotometryFilters.c.effective_wavelength,
+).pandas()
+null_values = pd.isnull(t["ucd"])
+df = t[null_values].sort_values(by=["effective_wavelength"])
+df_new = df.drop(columns=["effective_wavelength"])
+data_dicts = df_new.to_dict(orient="records")
```
dr-rodriguez, line 19: While this is fine, you may have been able to do the same with:
```
t = db.query(
    db.PhotometryFilters.c.band,
    db.PhotometryFilters.c.ucd,
    db.PhotometryFilters.c.effective_wavelength,
)
.filter(db.PhotometryFilters.c.ucd.is_(None))
.order_by(db.PhotometryFilters.c.effective_wavelength)
.pandas()
```

tests/test_photometry.py
```diff
@@ -0,0 +1,177 @@
...
+            )
+        )
+        .astropy()
+    )
+    assert len(t) == 969, f"found {len(t)} Y photometry entries"
+
+
+def test_photometry_filters_ucds(db):
+    # UCDS based on UCD1+ controlled vocabulary version 1.5
+    # https://www.ivoa.net/documents/UCD1+/20230125/index.html
```
dr-rodriguez, line 112: Would it be better if we use astropy's ucd checker? https://docs.astropy.org/en/stable/api/astropy.io.votable.ucd.check_ucd.html#astropy.io.votable.ucd.check_ucd

dr-rodriguez: If so, this may be an argument to also having this be part of the SIMPLE schema validation functions

scripts/updates/update_phot_filters.py
kelle: I couldn't figure out the NONE!!!  Is that construction documented somewhere?  In SQLAlchemey docs or someplace else?

tests/test_photometry.py
kelle: oh, neat!  Where would we put the "schema validation functions"? Is this that new thing in the schema.py?

dr-rodriguez: Yep, they go in schema.py. We've started testing that with the template db: https://github.com/astrodbtoolkit/astrodb-template-db/blob/main/schema/schema_template.py#L109 but it would be good to migrate that into SIMPLE as well

scripts/updates/update_phot_filters.py
dr-rodriguez: It's likely in the SQLAlchemy docs, but those can be hard to navigate sometimes. I ended up looking at I think the integrity tests since we had done similar queries before.

dr-rodriguez (approved):

tests/test_photometry.py
kelle: I tried to implement this but I can't figure out how to test it.  I can't get it to raise the ValueError.

dr-rodriguez: I wouldn't worry about implementing it in schema.py at this stage- we can do that on a dedicated PR and I expect we may want to do it in steps (eg, table by table). I anticipate we'll discover some issues with the data as we work through it.

dr-rodriguez: Also, if you mean why passing a wrong value doesn't return an error- I think we must use the ORM representation, simple inserts/modifications using direct sql will not trigger the validation functions.

kelle: well, I added it. Doesn't seem to impact anything. I'm ok with leaving it and the test for the moment.

dr-rodriguez: For the future (we may want a test_schema), this is how you could test/verify:
```
from simple.schema import PhotometryFilters
obj = PhotometryFilters(band="fake.fake", effective_wavelength=22, ucd="BAD")
```
That throws `ValueError: UCD BAD not in controlled vocabulary`

PR #496, opened 2024-05-01
Add instructions for how to use `spectrum_plottable`

kelle:
Just adding some instructions to the documentation of the Spectra table for using the `spectrum_plottable` function.


documentation/Spectra.md
```diff
@@ -34,4 +34,12 @@ For example: `$ASTRODB_SPECTRA/infrared/filename.fits`
...
    One of the dates should be used for the *observation_date*.
- - Spectra for companions should be associated with individual sources and not grouped with the primary source.
\ No newline at end of file
+ - Spectra for companions should be associated with individual sources and not grouped with the primary source.
+
+# Check if spectra are plottable by the website
+   ```Python
+   from simple.utils.spectra import spectrum_plottable
+   file = <path to file>
+   spectrum_plottable(file, show_plot=True)
```
dr-rodriguez, line 43: Maybe save the output and print it in a new line so users know that the return from this function is True/False as to whether the spectrum can be converted to a Spectrum1D object and be plotted. The function name could be clearer, but that can be a separate PR (eg, I would have recommended the function be called something like `is_spectrum_plottable`).

kelle: I don't mind renaming it to `is_spectrum_plottable`!

kelle:
Moved to https://github.com/astrodbtoolkit/astrodb_utils/pull/146

PR #505, opened 2024-06-11
wrote script for radial velocities

Exu-112:
Created csv with only sources with rv and rv_ref. Wrote script ingesting them into the RadialVelocities table.


scripts/ingests/ultracool_sheet/Ingest_radial_velocities.py
```diff
@@ -0,0 +1,63 @@
...
+
+    if len(match) == 1:
+        simple_source = match[0]
+        logger.info(f"Match found for {uc_sheet_name}: {simple_source}")
+        print(f"Match found for {uc_sheet_name}: {simple_source}")
+        rv_data = [{"source": simple_source, "radial_velocity_km_s": source["rv_lit"], "reference": source["ref_rv_lit"]}]
+        try:
+            with db.engine.connect() as conn:
+                conn.execute(db.RadialVelocities.insert().values(rv_data))
+                conn.commit()
```
kelle, line 57: This looks great!  However, we're trying to shift to a new method which I think looks like this. @dr-rodriguez , could you take a look also.

```suggestion
        rv_data = [{"source": simple_source, "radial_velocity_km_s": source["rv_lit"], "reference": source["ref_rv_lit"]}]
        rv_obj = RadialVelocities(**rv_data)
        try:
            with db.engine.session() as session:
                session.add(rv_obj)
                session.commit()
```

kelle (commented):
This looks great!  I'm just not sure about the way we want to actually do the ingest. I've asked @dr-rodriguez  to take a look.

kelle:
Just curious, how did you create the csv files with only rv data?  And I assume we only need the file that includes references, right?

kelle:
Also just noticed that uncertainty is missing.  Please also ingest the uncertainty on the measurement.

Exu-112:
I created the csv files manually using a bit of google sheets automation - figured it would be quicker than writing code for it. It doesn't seem that the new method for ingestion works on my side, though.

dr-rodriguez:
> I created the csv files manually using a bit of google sheets automation - figured it would be quicker than writing code for it. It doesn't seem that the new method for ingestion works on my side, though.

What do you mean that "the new method for ingestion" doesn't work on your side? Can you make the comment in the lines of code that you are referring to? I see over a hundred json files modified with radial velocities so it looks like it works. Is this just an old comment and it's working now?

scripts/ingests/ultracool_sheet/Ingest_radial_velocities.py
kelle: @dr-rodriguez Here's the convo with the "old" vs "new" way.

dr-rodriguez: Ah, `rv_data` must be a dictionary, not a list. If that's exactly as it was run, it could be why it failed. So remove the `[]` part of it.

__init__.py
dr-rodriguez, line 1: I'm curious as to the need for this extra `__init__.py` file. Did you mean to add it or was it done automatically? Did things not run without it?

Exu-112: I added it when trying to debug an error with my local python environment, it is not needed for the radial velocity script to run.

PR #511, opened 2024-06-14
ingest radial velocities from the Ultracool sheet

### LINKED ISSUE #506: Ingest radial velocities from the Ultracool Sheet
Part of #452 . [https://bit.ly/ultracoolsheet](https://bit.ly/ultracoolsheet?fbclid=IwAR1gDNgHgZKj_luFBdNqPXFxRrojTT2BwPFUfXSixscIQtHcrfcO2Q0F0uI
)
For right now, just ingesting the RVs for the sources already in the database.

Started in #511  by @exu-112.

Exu-112:
Short description: Ingesting radial velocities from the Ultracool sheet

Link to relevant issue: Closes #506

For data ingests:

- [x] Add new tests


Exu-112:
@dr-rodriguez tests pass locally, please review script before I upload json files

scripts/ingests/ultracool_sheet/Ingest_radial_velocities.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if len(match) == 1:
+        #1 Match found. INGEST!
+        simple_source = match[0]
+        if(uc_sheet_name in bad_sources):
+            continue
+        logger.info(f"Match found for {uc_sheet_name}: {simple_source}")
+        print(f"Match found for {uc_sheet_name}: {simple_source}")
+        rv_data = {"source": simple_source, "radial_velocity_km_s": source["rv_lit"], "radial_velocity_error_km_s": source["rverr_lit"],"reference": uc_ref_to_simple_ref(source["ref_rv_lit"])}
+        rv_obj = RadialVelocities(**rv_data)
```
dr-rodriguez, line 95: I would recommend setting rv_obj inside the try..except section. I don't remember if we've added explicit validation there, but if/when we do then creating that object can raise an error if any of the values are invalid. We can have another except handler (for something like ValueError) to catch these and inform the user.

scripts/ingests/ultracool_sheet/Ingest_radial_velocities.py
```diff
@@ -0,0 +1,112 @@
...
+    t=(
+        db.query(db.Publications)
+        .filter(db.Publications.c.bibcode==uc_ref_to_ADS[ref])
+        .astropy()
+    )
+    if(len(t)==0):
+        ingest_publication(db,bibcode=uc_ref_to_ADS[ref],reference=ref)
+        return ref
+    else:
+        return t["reference"][0]
```
dr-rodriguez, line 68: At first, I didn't see where this function was being called. Maybe a comment/docstring about what this function does and where it will be used would be useful.

dr-rodriguez:
I think the script and test looks good. I added a few minor comments. In general, I think some formatting could help readability- more spaces or line breaks. But that's something an autoformatter like ruff/black/darker can take care of.

For now, I think this is good and you can generate and add the JSON files and update the Versions table.

dr-rodriguez (approved):

kelle:
I am ok with not updating the versions table for this PR since 1) I'm at a conference this week and 2) We've got the Spitzer photometry ingest hot on the heels of this ingest.

scripts/ingests/ultracool_sheet/Ingest_radial_velocities.py
kelle: - Please move the `rv_obj` definition into the `try` as David suggest. 
- Also make a `ingest_radial_velocity` function.

kelle (approved):
This looks great!

PR #513, opened 2024-06-14
Improve `compile_header` function

kelle:
Improvements to make the `compile_header` function more user friendly.

### DIFF
--- a/simple/utils/spectra_convert.py
+++ b/simple/utils/spectra_convert.py
@@ -1,17 +1,20 @@
 import numpy as np
 import logging
-from pandas import to_datetime
+# from pandas import to_datetime
 from astropy.time import Time
 from datetime import date
 import astropy.io.fits as fits
 from astropy.table import Table
 import astropy.units as u
+import dateparser
+from astrodb_utils.photometry import assign_ucd
 
 logger = logging.getLogger("SIMPLE")
 
 
 def compile_header(wavelength_data, **original_header_dict):
     """Creates a header from a dictionary of values.
+    TODO: Check that RA and Dec are in degrees
 
     wavelength_data: an array of wavelengths
 
@@ -41,13 +44,13 @@ def compile_header(wavelength_data, **original_header_dict):
         "observatory",
     ]
 
-    keywords_given = list(original_header_dict.keys())
+    # keywords_given = list(original_header_dict.keys())
 
-    for key in keywords_given:
-        if key not in required_keywords:
-            raise Exception(f"Not expecting keyword: {key}. Add manually.")
-        else:
-            pass
+    # for key in keywords_given:
+    #    if key not in required_keywords:
+    #        raise Exception(f"Not expecting keyword: {key}. Add manually.")
+    #    else:
+    #        pass
 
     history = original_header_dict["history"]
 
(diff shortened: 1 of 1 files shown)


simple/utils/spectra_convert.py
```diff
@@ -56,81 +59,59 @@ def compile_header(wavelength_data, **original_header_dict):
...
-        time = None
-        logging.warning("No time")
-
-    try:
-        exposure_time = original_header_dict["exposure_time"]
-        header.set("TELAPSE", exposure_time, "[s] Total elapsed time")
-    except KeyError:
-        exposure_time = None
-        logging.warning("No exposure time")
+        object_name = input("REQUIRED: Please input a name for the object: ")
```
dr-rodriguez, line 73: I'm not a fan of the input function for several reasons, but here are two main points:

1. Testing this function is now more complex- you have to simulate a variable number of user inputs and while there are surely methods in pytest to do so, it'll require more setup.
2. The flow of this function is not at all consistent. If values exist, no inputs are needed, sometimes some inputs are needed but not others. Automating this, if you want to loop through a list, will become nearly impossible.

The original implementation, which I see you commented out around line 50 is better (though can be tweaked)- identify all missing keywords from your input dictionary and fail if any of them are missing.

simple/utils/spectra_convert.py
```diff
@@ -139,50 +120,75 @@ def compile_header(wavelength_data, **original_header_dict):
...
+            header.set("APERTURE", aperture, "[arcsec] slit width")
 
     # Publication Information
     try:
         title = original_header_dict["title"]  # trim so header wraps nicely
         header.set("TITLE", title, "Data set title")
     except KeyError:
-        pass
-        logging.warning("No title")
+        title = None
```
dr-rodriguez, line 184: As an alternative to so many try..except blocks, consider using something like:
```python
title = original_header_dict.get("title")
if title is not None:
   header.set("TITLE", title, "Data set title")
```
The python dictionary's `get` method automatically provides a default value of `None` if the key is not present: https://docs.python.org/3.10/library/stdtypes.html#dict.get

simple/utils/spectra_convert.py
kelle: The target user of this function (at the moment) is a high school student or undergrad who is using the spectrum for a different purpose. I want a function which will hold their hand through putting together our highest priority header keywords. 
- I have used the `input()`  function for what I consider the highest priority keywords.
- For keywords which they can likely figure out, e.g., Instrument , I've also used the `input()` function but labeled them as OPTIONAL.
- for keywords which may have existed in the original header, I've just pulled them from the original header with no user interaction. 

If you have another suggestion for how to do the hand holding, I'm open!

I also wasn't planning on doing too much testing in our CI...I was gonna let the students do the testing. As they find things which don't work, we can modify the function and add tests? 

I also recognize that this function is not mission critical to SIMPLE -- the database will still work with heterogenous spectra files -- so I'm ok if this function is not to the same standard as our ingest code.

kelle: all that being said, this could be a useful script in `astrodb_utils` at some point, especially since it's striving to be consistent with SpectrumDM 1.2.  We could rename it `compile_ivoa_spectrum_dm_header`.  And I'm working on doing the same thing for light curves at the minute.  So maybe making it high quality code is a good idea....

dr-rodriguez: One thing that occurs to me is that we can split this into two (or more) functions: 
1. one function that takes a dictionary and compiles the header and 
2. one that validates and creates the dictionary

That later function can be one using the `input()` calls (and we can add helper functions if needed), but the compile header function is more streamlined and can be tested or automated more readily.

kelle: I think two functions is a very good idea. I think I may also go ahead and develop this in `astrodb_utils` instead of here in SIMPLE.

PR #539, opened 2024-07-16
refactor ingest_spectral_types function

### LINKED ISSUE #481: Refactor ingest_spectral_types function
Should be just for one spectral type, not for a list of them.

Change:
```python
n_added = 0
n_skipped = 0
```
to a `flags` dictionary with `added` and `skipped` keys.
```python
flags = {'added': 0, 'skipped': 0}
if <added>:
    flags['added'] = 1
if <skipped>:
     flags['skipped]=1
...
return flags

### PR DESCRIPTION
**Pull request by Exu-112**
Short description: refactor ingest_spectral_types function to ingest one at a time and use the new ingest method

Link to relevant issue: Closes #481

### CODE REVIEW

**Comment by kelle** [comment]
Please take a look. If all is okie dokie, I'll add the 3863 modified JSON files with the new photometric column.

File: documentation/SpectralTypes.md [hunk 1]
```diff
@@ -20,4 +21,6 @@ Spectral Type Codes:
  - 69 = M9
  - 70 = L0 
  - 80 = T0
- - 90 = Y0
\ No newline at end of file
+ - 90 = Y0
+
+ Regime is required however, regime = "unkwown" can be used when the regime is unknown.
```
dr-rodriguez, line 26: Typo: "unknown"

simple/utils/spectral_types.py
```diff
@@ -3,320 +3,317 @@
...
+                logger.debug(
+                    "No spectral type error found, setting adopted flag to True"
+                )
         else:
-            msg = "Unexpected state"
-            logger.error(msg)
-            raise RuntimeError
+            adopted = True
+            old_adopted = None
+            logger.debug("No adopted data found, setting adopted flag to True")
```
dr-rodriguez, line 134: I'm OK with this for now BUT this screams for a helper function that solely handles the determination of whether adopted should be set or not. As a helper function, it can be independently tested and implemented anywhere we have adopted columns.

simple/utils/spectral_types.py
```diff
@@ -3,320 +3,317 @@
...
+    if len(db_name) != 1:
+        msg = f"No unique source match for {source} in the database "
+        raise AstroDBError(msg)
+    else:
+        db_name = db_name[0]
+
+    source_spt_data = (
+        db.query(db.SpectralTypes).filter(db.SpectralTypes.c.source == db_name).table()
+    )
+    logger.debug(f"Pre-existing Spectral Type data: \n {source_spt_data}")
```
dr-rodriguez, line 79: The function would flow better if lines 76-79 were after line 102. Essentially, here you are defining `source_spt_data`, then you break to check for duplicates, then you go back to analyzing `source_spt_data`. Why not have the definition right before you use it? From what I can tell the duplicate check doesn't require this.

simple/utils/spectral_types.py
dr-rodriguez: And to clarify- not updating the database, just determining and returning the values of adopted/old_adopted.

simple/utils/spectral_types.py
```diff
@@ -3,320 +3,317 @@
...
+            msg = f"Spectral type ingest failed with error {e}\n"
+            if raise_error:
                 logger.error(msg)
                 raise AstroDBError(msg)
+            else:
+                logger.warning(msg)
 
-    msg = f"Spectral types added: {n_added} \n" f"Spectral Types skipped: {n_skipped}"
-    logger.info(msg)
+    # check that adopted flag is successfully changed
```
dr-rodriguez, line 201: Should these checks/validation just be part of the tests? In fact, I'm pretty sure we already have tests for this. That would make this function simpler to parse and unit test.

dr-rodriguez: If you want to have it in spectral_type.py, they should be its own separate function (which can be called by `ingest_spectral_types` if you really need to).

tests/test_spectral_type_utils.py
dr-rodriguez, line 1: The first 3 tests here can benefit from using `pytest.mark.parametrize` with some rewriting. That allows essentially to run multiple instances of the same test with different inputs without them blocking one another. It also results in cleaner code with less repetition. We've used that before in other examples.

dr-rodriguez:
Overall, this gets the job done. I think we can do better in terms of the function flow and styling. One of the things I see a lot of these days is giant functions that do more than one thing, which is bad design and hard to test. One way we could work towards catching that is implementing a linter to this project. I'm using ruff with some custom settings in one of my own projects and the McCabe complexity indicator is useful for this: https://docs.astral.sh/ruff/rules/complex-structure/ 

I recommend creating a few issues from the comments above and then we can approve and update the json files.

simple/utils/spectral_types.py
kelle: I at least made a helper function just for spectral type. Every adopted column is gonna have different logic so making a generic one will be hard.

kelle:
Excellent suggestions. I implemented most of them.

simple/utils/spectral_types.py
dr-rodriguez: Yep, this is good for now. I was thinking that it would take as input `source_spt_data` so it wouldn't need to know how to query the database (or do any DB access at all), but since there are some other if-statements relevant to spectral types it's worth tackling this on its own.
I created an issue for a more generalized set of functions: https://github.com/SIMPLE-AstroDB/SIMPLE-db/issues/542

simple/utils/spectral_types.py
```diff
@@ -3,320 +3,334 @@
...
+
+    # Numeric part of type
+    format = f".{decimals}f"
+    spt_type = f"{spt_type}{spectral_code% 10:{format}}"
+    logger.debug(f"Converting: {spectral_code} -> {spt_type}")
+
+    return spt_type
+
+
+def adopt_spectral_type(db, source, spectral_type_error):
```
dr-rodriguez, line 234: These new adoption functions are un-tested, but I think we can take care of that when we address https://github.com/SIMPLE-AstroDB/SIMPLE-db/issues/542

dr-rodriguez (approved):

simple/utils/spectral_types.py
kelle: added #TODOs in the test file

PR #541, opened 2024-07-23
Restructured to have reference JSON directory

dr-rodriguez:
Restructured the database with the changes from https://github.com/astrodbtoolkit/AstrodbKit2/pull/77 
The tests are likely to fail until that gets merged and released.

The core change is that the `reference_directory` is now an optional variable that can be set (defaults to "reference") when saving/loading the database. Anything set as a reference table will get stored in that sub-directory. This is backwards compatible in the sense that if the reference tables are not in that directory, it will check the original (parent) directory.


dr-rodriguez:
I noticed that github will suppress anything over 3000 files, but this particular PR is relatively simple as it is just moving data from the top-level data directory into data/source and data/reference. Once astrodbkit2 is updated, we can re-run the tests and confirm everything is working well (locally on my machine all tests pass with the new astrodbkit2).

kelle (approved):
AMAZING!!!!

PR #548, opened 2024-07-31
Use `check_spectrum_plottable` from `astrodb_utils`

kelle:
The `spectrum_plottable` function has been incorporated into `astrodb_utils` and should not be in this repo anymore. The changes made there necissated some changes in the `ingest_spectrum` function.

### DIFF
--- a/tests/test_spectra_utils.py
+++ b/tests/test_spectra_utils.py
@@ -7,83 +7,111 @@ from astrodb_utils.utils import (
 from simple.utils.spectra import (
     ingest_spectrum,
     # ingest_spectrum_from_fits,
-    spectrum_plottable,
 )
 
 
-@pytest.mark.filterwarnings(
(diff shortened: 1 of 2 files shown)


tests/test_spectra_utils.py
```diff
@@ -7,83 +7,111 @@
...
 )
-@pytest.mark.filterwarnings(
-    "ignore",
-    message=(
-        ".*'erg/cm2/s/A' contains multiple slashes, "
-        "which is discouraged by the FITS standard.*",
-    ),
+@pytest.mark.parametrize(
+    "test_input, message",
+    [
```
dr-rodriguez, line 22: As a note, if you want to prevent auto-formatters from reformatting text (for example if you prefer something more compact even if it's not Black/ruff/darker-standard) you can preface the block with `# fmt: off` and then end it with `# fmt: on`

dr-rodriguez (approved):
Looks good to me.

PR #564, opened 2024-11-02
Fix `scheduled_checks` test

kelle:
The scheduled test which tests all the URLs once a month is not passing. I'm trying to fix it.


kelle:
Oh, fascinating. It was broken on my FORK!  Cause I don't keep my main up-to-date. How to prevent scheduled_checks from running on forks, I wonder?

PR #605, opened 2025-04-29
BONES Archive Ingest Spectral Types

### LINKED ISSUE #544: Ingest the BONES Archive of low metallicity sources
The [BONES Archive of low metallicity sources](https://docs.google.com/spreadsheets/d/1b9SUUshady_efTJaWEXiy4g-1n-JOzG_Mcy6j1CFE8k/edit?gid=0#gid=0). 

Compiled by @huntbrooks85.

From the sheet:
- [x] sources. Done in #590 
- [ ] spectral types. #605 
- [ ] spectra

Directly from the catalogs:
- [ ] Gaia photometry and astrometry 
- [ ] SDSS, PanSTARRs, UKIDSS, 2MASS, WISE, AllWise, UKIDSS, etc photometry

lesliech1004:
Short description: ingest spectral types from spectra sheet from the BONES archive

Link to relevant issue: Closes #544 (spectral types only)

For data ingests:
- [ ] includes script used for ingest
- [ ] includes modified JSON files 
- [ ] Add new tests


kelle:
This looks GREAT!  Great job ingesting the publications and having multiple counters!

The only thing that I'm seeing as a problem is that some of the spectral types which were not ingested because they are already in the database actually have different spectral types. For example,WISEA J000430.66-260402.3 is the database as "T2 (blue)" but in the Bones archive as "sdT2".  So, I need to modify the ingest script and/or the schema to see this situation as *not* a duplicate.

scripts/ingests/BONES_archive/ingest_BONES_spectral_types.py
```diff
@@ -94,15 +92,17 @@ def extractADS(link):
...
-                pub_ingested +=1
-        except:
-            logger.warning("Find and ingest pattern didn't work" + ads)
-        ref = find_publication(db=db, bibcode=ads)
-        #ingest the spectral types
+                pub_ingested += 1
+                ref = find_publication(db=db, bibcode=ads)
+            except Exception as e:
+                logger.error("Find and ingest pattern didn't work" + ads)
+                raise e
```
kelle, line 103: When you encounter an exception, you should raise the error so that the code stops and you can figure out what's wrong.

scripts/ingests/BONES_archive/ingest_BONES_spectral_types.py
```diff
@@ -117,7 +117,10 @@ def extractADS(link):
         logger.warning(msg)
         if "Spectral type already in the database" in str(e):
             already_exists+=1
-        skipped+=1
+        else:
+            skipped += 1
+            raise e
```
kelle, line 122: This is good -- you are expecting some to be already in the database, BUT there might be some other types of exceptions. When you encounter those, the exception should be raised, in this case, with `raise e`

scripts/ingests/BONES_archive/ingest_BONES_spectral_types.py
```diff
@@ -6,24 +9,19 @@
...
 
-sys.path.append(".")
-from astropy.io import ascii
+astrodb_utils_logger = logging.getLogger("astrodb_utils")
+logger = logging.getLogger(
+    "astrodb_utils.bones_sptypes"
+)  # Sets up a child of the "astrodb_utils" logger
 
+logger.setLevel(logging.INFO)  # Set logger to INFO/DEBUG/WARNING/ERROR/CRITICAL level
+astrodb_utils_logger.setLevel(logging.INFO)
```
kelle, line 22: I'm finally getting to understand how logging works!

kelle (changes requested):
You script doesn't ingest at least one source because the `extract_ADS` function didn't work. I suspect that you didn't see this because you weren't raising the exceptions in your try/except pairs.  Run the script now and notice that the code stops when it encounters the error with the publication rather than just skipping it and moving on.  Modify your ingest script to add this publication.

kelle:
This looks great!  Please push the JSON files and modified tests!

kelle (approved):
Great work!  Gonna merge this.

Issue #529, opened 2024-07-08
Data ingest request from: Zhang et al. 2018


jgizis:
Article title: "Primeval very low-mass stars and brown dwarfs – III. The halo transitional brown dwarfs"

https://academic.oup.com/mnras/article/479/1/1383/5001885

the supplementary data on the website includes X-Shooter visible and IR spectra of L subdwarfs.

kelle:
or @zjzhang42, want to help?

jgizis:
Ok, I will confess that I submitted this because my super-naive attempt to read the fits files with specutils failed, so I can't do that part.  But I am happy to populate a table with the fitted parameters from this and the other papers in the series.  

(Although I was able to read the supplement file in 2019MNRAS.486.1840Z with no problem.).

jgizis:
Oh, looking at the supplemental files I think Paper IV has supplemental files that include many of the spectra from the earlier papers. 

https://academic.oup.com/mnras/article/480/4/5447/5062823

zjzhang42:
@kelle  Hey, just FYI I am not the first author - Zenghua Zhang.

kelle:
Ah, apologies for my confusion!

kelle:
Here's a sheet to fill out for the missing spectra: [SIMPLE Spectra Ingest Template](https://docs.google.com/spreadsheets/d/1EOaIV_NPnow8jnRpi_086zwP8vXh2efQcYcnPdqmq-E/edit?gid=0#gid=0)

And for the modeled parameters: [SIMPLE Modeled Parameters Template](https://docs.google.com/spreadsheets/d/1NsfTozaSDUiaM-uE4kJVs42KXc_n3Uj1Q4UtqCgswVE/edit?gid=0#gid=0)

Are all the sources already in or do they need to be added as well?

kelle:
It looks like symmetric uncertainties are provided for all of the modeled parameters. If that's not the case, let's discuss more. There's a plan to deal with asymmetric uncertainties but it's not implemented yet (#484 ).

jgizis:
ok.  I think the sources are all already in but if any are not I'll let you know.

kelle:
Most of these data need to be smoothed significantly in order to see the shape of the spectrum. I'm thinking we should go ahead and ingest the smoothed data so that they will be worth looking at on the website.  The original data will still be available as the "original_spectrum".  Objections?  I would also add a comment.

PR #630, opened 2025-06-23
Ingest Zhang18 spectra from FITS file

ying2212:
Short description: Include what type of data being ingested and appropriate references.

Link to relevant issue: #529 

For data ingests:
- [ ] includes script used for ingest
- [ ] includes modified JSON files
- [ ] Add new tests


kelle:
Great start with converting the spectra!  My one suggestion at this point is to do the various instruments/modes separately.  Right now, it looks like the wavelength units for "Xshooter_VIS" isn't quite right but "OSIRIS" looks good.

kelle:
The spectra are looking great!  But it looks like there are still some spectra which are failing or being skipped. We need to figure out what's going on with those.

ying2212:
I’m running into errors with ingesting `access_url` in the `Spectra` table, seems like it only accepts URL format, but I’m working with local FITS files. I asked Carolina, and she said SIMPLE bdnyc has cloud storage for the files. Should I pull them from there instead?

kelle:
We decided to just make two spectra, one binned and unbinned.  New question: how to deal with duplicate metadata and primary keys?

PR #663, opened 2025-08-20
Create SIMPLE Demos

canavarrete01:
Short description: Moved from secret simple, two notebooks for querying the database and another for visualizing spectra and data.


canavarrete01:
@kelle QueryDemo.ipynb needs comments / notes on how to further improve the notebook!

QueryDemo.ipynb
```diff
@@ -0,0 +1,3670 @@
...
+       " ...]"
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "db.query(db.Sources).all()    "
```
dr-rodriguez, line 1159: To avoid so many results consider adding a limit, may also want to output as a table for easier viewing:
`db.query(db.Sources).limit(10).table()`

Even if you don't apply a limit, I think the notebook's rendering of astropy tables will be better than the default tuple output.

QueryDemo.ipynb
```diff
@@ -0,0 +1,3670 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "data = db.inventory('WISEPC J033349.34-585618.7', pretty_print=True)\n",
+    "print(data) "
```
dr-rodriguez, line 1459: You don't need to add `print(data)` if you already set `pretty_print=True`.    
If you notice, this causes the dictionary to be printed twice, once in a fairly easy to read way and the other as a big blob.

QueryDemo.ipynb
```diff
@@ -0,0 +1,3670 @@
...
+      ]
+     },
+     "execution_count": 10,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "db.search_string('Cruz18', fuzzy_search=False)  # search for strings exactly matching Cruz19 anywhere in the database\n",
+    "db.search_string('Cruz18', fuzzy_search=False, fmt='pandas')  # as above, but have each table as a pandas dataframe"
```
dr-rodriguez, line 1777: I don't think we need to show both results, maybe just the one with the `fmt` parameter passed in.

QueryDemo.ipynb
```diff
@@ -0,0 +1,3670 @@
...
+    }
+   ],
+   "source": [
+    "# An example of query wth an empty table\n",
+    "incorrect_results = db.query_region(SkyCoord(144.3, 29.5, frame='icrs', unit='deg'), output_table='Sources', ra_col='ra', dec_col='dec')  # specifying the name of the table with coordinate information\n",
+    "print(incorrect_results)  \n",
+    "\n",
+    "# An example that queries the database correctly \n",
+    "correct_results = db.query_region(SkyCoord(144.395292, 29.528028, frame='icrs', unit='deg'), output_table='Sources', ra_col='ra', dec_col='dec')  # specifying the name of the table with coordinate information\n",
+    "correct_results"
```
dr-rodriguez, line 1831: You may want to rework your outputs here. The end result is a table but you've printed both the incorrect and (implicitly) the correct results. So which of them is it? Better to print out a string with some text saying something along the lines of Number of incorrect results: len(incorrect_results) or such to be clear.

QueryDemo.ipynb
```diff
@@ -0,0 +1,3670 @@
...
+       " ('51 Eri b', 69.40041, -2.47361, None, None, '0437-0228b', 'Maci15', None, None),\n",
+       " ('2MASS J21144103-4339531', 318.6713, -43.6652, None, '2000', None, 'Gagn15.33', None, None),\n",
+       " ('2MASS J12321827-0951502', 188.0761245, -9.863944, None, None, '1232-0951', 'Reid08.1290', None, None),\n",
+       " ('WISEA J011154.36-505343.2', 17.9765, -50.8954, None, '2000', None, 'Luhm14.126', None, None),\n",
+       " ('ULAS J133836.97-022910.7', 204.6543, -2.4854, None, '2000', None, 'Zhan17.3040', None, None),\n",
+       " ('2MASSI J1658037+702701', 254.51585415, 70.45043333, None, None, '1658+7026', 'Gizi00', None, None),\n",
+       " ('CFBDS J022644.65-062522.0', 36.686, -6.4228, None, '2000', None, 'Albe11', None, None),\n",
+       " ('2MASS J05574102-1333264', 89.420954, -13.557334, None, None, None, 'Cruz18', None, 'not M'),\n",
+       " ('2MASS J09393548-2448279', 144.897834, -24.80775, None, None, '0939-2448', 'Tinn05.2326', None, None),\n",
+       " ('2MASS J22590210+1156035', 344.758944, 11.933929, None, None, None, 'Zhan18.2054', None, 'Discovery reference from t
```
dr-rodriguez, line 2901: If you wanted to demonstrate the use of `sql_query` that's fine, but note that this can be more simply done with: `db.query(db.Sources).count()`

dr-rodriguez, line 2933: `sql_query` can also take a `fmt` parameter if you want to work with astropy tables or pandas dataframes.

dr-rodriguez, line 3140: I would not have this cell. You've shown how to use the Astrodbkit query methods and how to use sql_query, but now you're going into low-level API for SQLAlchemy itself. While useful, it probably doesn't belong in a notebook demonstrating quick/useful queries of SIMPLE.

QueryDemo.ipynb
dr-rodriguez, line 1: Overall, think about the purpose of this notebook, what you want to get across and focus on that. It helps to add a lot more text, both in markdown and python to guide the user.

kelle (changes requested):
The new flow in the spectra notebook is good!  It should stop after the nicely formatted figure with the three spectra.  Everything after "Now lets try plotting spectra from multiple sources using a couple different metrics we can decide on within SIMPLE" should go into a separate notebook.

Demo_Queries_1.0.ipynb
kelle, line 1: Title: Querying the database using Python

kelle (approved):
Great work!

PR #671, opened 2026-03-03
Ingest Zhang 2018

ying2212:
Short description: Fix simple.yaml to add mode as PK in spectra table and JSON ingestion for Zhang18.

Link to relevant issue: Closes #

For data ingests:
- [x] Add new tests


data/reference/Instruments.json
```diff
@@ -516,5 +516,26 @@
...
+    {
+        "instrument": "Xshooter",
+        "mode": "Echelle-Smoothed",
+        "telescope": "ESO VLT",
+        "description": null,
+        "reference": null
+    },
+    {
+        "instrument": "IMACS",
+        "mode": "Missing",
```
kelle: Can mode be null instead of "Missing"?

kelle: let's change these "Missing"s to "spectroscopy"

kelle:
We also need to double check the discovery references for these 14 new sources.

PR #683, opened 2026-04-14
Gaia Epoch Ingest JSON Files

### LINKED ISSUE #667: Get updated coords and epochs from Gaia
Get updated coords and epochs from Gaia.  We've already the cross match so this should be straightforward. It's just a modification of a script that's already written
https://github.com/SIMPLE-AstroDB/SIMPLE-db/tree/main/scripts/ingests/Gaia

Logan-Huang:
Short description: Adds the 1265 json files from the successful Gaia ingest. The other 597 modified json files had their adopted values set to booleans instead of 0/1s. The ingest script was edited to be less destructive, not setting values to None anymore. 

Link to relevant issue: Closes #667


kelle:
Exciting!!!

Next step is to modify the tests.

kelle:
Great!  Next step is to add some tests which verify the modifications that you made. Focus on the 1265 epochs.  The test should go at the bottom of this module: https://github.com/SIMPLE-AstroDB/SIMPLE-db/blob/main/tests/test_data_astrometry.py.  Reading the other tests in that module should give you inspiration for how to write a new test that verifies the number of sources with and without an epoch.

kelle:
I also have to say, having the 587 other modified files with the boolean fix is very distracting.  Is it possible to open a new PR that *only* does the epoch ingest.  Save the boolean fix for it's own PR?

kelle:
In order to do that, you'd basically start over and re-run your script. Copy your script somewhere temporary (like your desktop), make a new fresh branch from an updated main, move your script into the new branch, modify the script to not touch the booleans (but save this code for later), run the script and save the JSONs, open a new PR with only 1265 modified JSON files.

